PR #2, opened 2018-05-21
Add internal config package

matthewrsj:
The internal config package adds a ReadConfig function which reads
configuration files from the system. The configuration files are read in
order starting with the passed in path and ending with the default
system configuration path. If no configuration file is found the
compiled defaults are used. The files are read in the following order
with the first file found being used.

Passed configuration file path, if defined
$HOME/.config/diva/config.toml
/etc/diva/config.toml
/usr/share/defaults/diva/config.toml


internal/config/config.go
```diff
@@ -0,0 +1,88 @@
...
+// The defaultConfig is expected to supply a complete set of default values
+// while the systemConfig and userConfig may define any subset of the total
+// config values. The values defined in systemConfig overwrite the
+// corresponding values in defaultConfig. The values defined in defaultConfig
+// overwrite the corresponding values in systemConfig and defaultConfig.
+func ReadConfig() (*config, error) {
+	var c config
+	// read default system config
+	_, err := toml.DecodeFile(defaultConfig, &c)
+	// it is always an error if the defaultConfig cannot be loaded
```
icarus-sparry: Didn't we suggest that we could have compiled in values as a final fallback, rather than requiring /usr/share/defaults/diva/config.toml?

From the point of view of testing, how do we override the system provisioned config files?

matthewrsj: Ok yeah I was thinking of that as I wrote that comment. I'll add that in.

> From the point of view of testing, how do we override the system provisioned config files?

By providing a configuration file under /etc or $HOME/.config? Or am I missing your point?

bryteise: No --config option to pass the file via cmdline?

matthewrsj: At the time of this PR the cmdline doesn't exist, but I can add an argument to `ReadConfig` I guess.

icarus-sparry: @matthewrsj You can add or replace options with this approach (each config file overrides what the previous one did), but you can't remove things unless you have an understanding that an empty string is the same as "not set".

matthewrsj: @icarus-sparry well true. That doesn't seem unreasonable to me though.

matthewrsj: And it means the user can just define what they want to override without having to provide the entire configuration file.

matthewrsj:
Added default config to code and path argument to ReadConfig

internal/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+// in defaultConf() will be used. If the configPath argument is not an empty
+// string only that path will be read for configuration values, with the
+// defaultConf() fallback values still in place. The configuration files are
+// read in the following order.
+//
+// defaultConfig "/usr/share/defaults/diva/config.toml"
+// systemConfig  "/etc/diva/config.toml"
+// userConfig    "$HOME/.config/diva/config.toml"
+//
+// Each config file may provide any subset of the total configuration values.
```
icarus-sparry: Whilst this works, I think it is better to take the following approach (code just typed into comment, not formatted, syntax might be wrong, ...)

     c:= defaultConf() // Sets up everything that is now expected, which allows adding new options which are not mentioned in the toml files
      for _,name := range []string{ configPath, userConfig, systemConfig, defaultConfig} {
                        if name != "" {      // handle no --config= on the command line
                                         _, err := toml.DecodeFile(name, &c) 
                                         if err == Error_file_does_not_exist { continue ; } // ignore missing file 
                                         return &c, err // i.e. return on first existing file, including unreadable file error
                      }
          }
          return &c, nil

What you lose is the ability to just do overrides, you have to have all the options in whichever config file the system ends up using. Against that you gain the ability to tell the difference between an option not being specified and it being specified but empty.  (There is a big difference between not having a bank account and having one but with no money in it). This makes it simpler to add new optional values to the config file. You also have a simpler processing model, `use the first config file specified` rather than `use the value in this config file unless it is overridden by a value in this file, unless that is overridden by a value in this file, unless it is overridden by a value in this file`.

I think it will make testing simpler as well, you just pass a config file as a command line parameter and then the state of the rest of the system will not affect anything.

matthewrsj: > What you lose is the ability to just do overrides, you have to have all the options in whichever config file the system ends up using.

Not true in your code. When you pass `&c` to `toml.DecodeFile` it only updates or adds what is present in `c`. Since the code here does not redefine `c` every time omitted values will keep their defaults.

Of course that can easily be changed by creating a new config every time.

Your argument to make it simpler is fair though, as is the argument to simplify testing. I'll do it your way. (my way was cooler just sayin 😝)

icarus-sparry: Sorry, I wasn't clear. `What you lose is the ability to just do overrides` was meant to refer to `~/.config/diva/config.toml` overriding `/etc/diva/config.toml`, rather than the compiled in defaults.

The point of the compiled in defaults is that they should make sense for **this** version of the program. (As an aside, typical go libraries are encouraged to make the zero value useful so you don't have to specify them). In contrast the values in the config.toml files need to make (at least some) sense for **every** version of the program.

I am not disputing your way is cooler, I just want something that is simpler to debug when things go wrong. 🍻

matthewrsj:
@icarus-sparry I followed your recommendation.

internal/config/config.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Config struct that defines the layout of the configuration file
+type Config struct {
+	Mixer mixConfig
+	Paths pathConfig
+}
+
+func defaultConf() Config {
+	ws := filepath.Join(os.Getenv("HOME"), "clearlinux")
```
phmccarty: Do we need to be concerned about `HOME` not being defined in the environment?

icarus-sparry:
Yes, we do need to be concerned about HOME not being defined, and we probably ought to worry about using $XDG_CONFIG_HOME as an override for $HOME/.config.

However lets get some code in now and worry about the edge cases in later pull requests.

phmccarty:
Okay, then I think we should add a FIXME (or open a new github issue) to track the issue.

PR #6, opened 2018-05-22
Add Makefile and .travis.yml for CI

matthewrsj:


matthewrsj:
\o/

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: go
+sudo: required
+
+go:
+    - 1.9
```
icarus-sparry: Are we not up to 1.10 these days?

Makefile
```diff
@@ -0,0 +1,93 @@
+# Makefile used to create packages for diva. It doesn't assume that the code is
```
icarus-sparry: It would be more go-like to get rid of this Makefile. However as a quick and dirty solution we can live with it.

.travis.yml
matthewrsj: hm we are indeed.

Makefile
matthewrsj: I like having the makefile to enable https://github.com/clearlinux/autospec for this project.

.travis.yml
matthewrsj: for some reason `go get` on gometalinter fails with go 1.10 in travis

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+
+go_import_path: github.com/clearlinux/mixer-tools
```
Gabib32: github.com/clearlinux/diva ?

matthewrsj: yikes, you can tell where I got this code from.

matthewrsj:
Added a version flag to the root command as well since the version needed to be defined in order for the make release target to work.

PR #11, opened 2018-05-29
helpers: Download and cache RPMs

matthewrsj:
Provide a helpers internal package that provides functionality to
download and cache RPMs from the c.UpstreamURL RPM repository. Parse
repo metadata to ensure we only download those RPMs in the repo. Do not
download RPMs that already exist under \<cache\>/rpms/\<version\>/packages.
Print a progress bar when stdout is in a TTY.


bryteise:
Did you prefer to avoid dnf reposync for this task?

matthewrsj:
@bryteise just a desire not to add DNF as a dependency.

internal/helpers/helpers.go
```diff
@@ -0,0 +1,117 @@
...
+func DownloadToFile(url, filename string) error {
+	resp, err := download(url)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	out, err := os.Create(filename)
```
icarus-sparry: Perhaps this should do more, in particular download to a temporary file in the same directory and then at the end do a rename? This will mean that if the process is aborted (ctrl-c for example) then you will not be left with a truncated file.

internal/helpers/helpers.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return nil
+}
+
+// StdoutIsTTY returns true if Stdout is to a TTY
+func StdoutIsTTY() bool {
+	_, _, err := syscall.Syscall6(
+		syscall.SYS_IOCTL,
+		os.Stdout.Fd(),
```
icarus-sparry: By definition os.Stdout.Fd() is 1.

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,233 @@
...
+	urlCh := make(chan string)
+	errorCh := make(chan error, workers)
+
+	dlWorker := func() {
+		for url := range urlCh {
+			base := filepath.Base(url)
+			outFile := filepath.Join(outPath, base)
+			// do not download again if it already exists
+			if _, err := os.Stat(outFile); err == nil {
+				if StdoutIsTTY() {
```
icarus-sparry: Needs a helper function to do bar.Add(1) if it is a tty.

Or you could restructure this to just use a channel to note the things that have been completed (I don't know if the bar package takes care of simultaneous Add(1) requests, but a channel does), and just have a go routine to drive the bar if it is a terminal and otherwise just be a data sink.

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,233 @@
...
+	baseURL := fmt.Sprintf("%s/releases/%d/clear/x86_64/os/Packages", c.UpstreamURL, version)
+	for _, p := range v.Packages {
+		rpmURL := fmt.Sprintf("%s/%s-%s-%s.%s.rpm", baseURL, p.Name, p.VR.V, p.VR.R, p.Arch)
+		packages = append(packages, rpmURL)
+	}
+
+	return packages, nil
+}
+
+func downloadAllRPMs(packages []string, workingDir string) error {
```
icarus-sparry: This can deadlock if there are too many download errors, for the same reasons that I explained last time.

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// GetUpstreamRepoFiles downloads all RPM packages from the RPM repo at
+// c.UpstreamURL by first paring the repo metadata. These packages are
+// downloaded to the c.CacheLocation/rpms/<version>/packages/ if they do not
+// already exist there.
+func GetUpstreamRepoFiles(version int) error {
+	var err error
+	c, err = config.ReadConfig("")
+	c.UpstreamURL = "https://clearnas.ostc.intel.com"
```
icarus-sparry: I thought we were  not allowed to mention internal hosts in public.

matthewrsj:
Comments from @icarus-sparry addressed, xz decompression support added (@phmccarty)

internal/helpers/getrepo.go
matthewrsj: Fixed.

icarus-sparry: Looks much better, thank you.

Now that you have implemented my prefered way of handling this stuff, I can confess that the easy fix would have been to change line 140 to be

              errorCh := make(chan error, len(packages))

i.e. to allow one error per package rather than one error per worker. This would also have fixed the potential deadlock. 😌

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,255 @@
...
+		for url := range urlCh {
+			base := filepath.Base(url)
+			outFile := filepath.Join(outPath, base)
+			// do not download again if it already exists
+			if _, err := os.Stat(outFile); err == nil {
+				progressIfTTY(bar)
+				continue
+			}
+			dlErr := Download(url, outFile)
+			if dlErr != nil {
```
icarus-sparry: You could restructure this to be

    for url := range {
            base :=
            outFile :=
            var err error
            if _, err = os.Stat(outFile) ; err != nil {
                        err = Download(url, outFile)
            }
            progressIfTTY(bar)
            if err != nil {
                     errorCh <- err
            }
      }

25% smaller, but probably not worth worrying about.

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,255 @@
...
+				errorCh <- dlErr
+				progressIfTTY(bar)
+				continue
+			}
+			progressIfTTY(bar)
+		}
+		wg.Done()
+	}
+
+	fmt.Printf("Downloading %d RPMs using %d workers\n", len(packages), workers)
```
icarus-sparry: Gut says either send to stderr or move  inside the IF statement to make it less chatty

internal/helpers/helpers.go
```diff
@@ -0,0 +1,186 @@
...
+	// download to file, no native xz compression library in Go
+	if err := Download(url, target+".xz"); err != nil {
+		return err
+	}
+
+	return RunCommandSilent("unxz", "-d", "-T", "0", target+".xz")
+}
+
+// StdoutIsTTY returns true if Stdout is to a TTY
+func StdoutIsTTY() bool {
```
icarus-sparry: Should cache the result so it doesn't need to do a syscall each time.

internal/helpers/helpers.go
```diff
@@ -0,0 +1,186 @@
...
+	_, _, err := syscall.Syscall6(
+		syscall.SYS_IOCTL,
+		1, // stdout file descriptor
+		syscall.TCGETS,
+		uintptr(unsafe.Pointer(&syscall.Termios{})),
+		0, 0, 0,
+	)
+	return err == 0
```
icarus-sparry: Strictly all this does is check that an ioctl with a magic value doesn't produce an error. However that is probably good enough.

internal/helpers/helpers.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// move tempfile to final now that everything else has succeeded
+	return os.Rename(tmpFile, filename)
+}
+
+// DownloadFile downloads from url and extracts the file if necessary using the
+// compression method indicated by the url file extension. If there is no file
+// extension or the extension does not match a supported compression method the
+// file is downloaded as-is.
```
icarus-sparry: We can probably do better by looking at the first few bytes in the file.

This would require a change, i.e. we download the file (perhaps to memory) and then see what to do with it.

I can't remember if a http.Response supports seek or not. If not then would need to wrap it in a buffio.Reader

internal/helpers/helpers.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	// write to a temporary file so if the process is aborted the user is
+	// not left with a truncated file
+	tmpFile := filepath.Join(filepath.Dir(filename), ".dl."+filepath.Base(filename))
```
icarus-sparry: 📯

internal/helpers/helpers.go
```diff
@@ -0,0 +1,206 @@
...
+	return nil
+}
+
+func xzExtractURL(url, target string) error {
+	// download to file, no native xz compression library in Go
+	if err := Download(url, target+".xz"); err != nil {
+		return err
+	}
+
+	return RunCommandSilent("unxz", "-d", "-T", "0", target+".xz")
```
phmccarty: No need for the `-d` option if you are running `unxz`. So either run `unxz` or `xz -d`.

internal/helpers/getrepo.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	return nil
+}
+
+// GetUpstreamRepoFiles downloads all RPM packages from the RPM repo at
+// c.UpstreamURL by first paring the repo metadata. These packages are
+// downloaded to the c.CacheLocation/rpms/<version>/packages/ if they do not
+// already exist there.
+func GetUpstreamRepoFiles(version int) error {
```
phmccarty: version should be a `uint`, perhaps?

internal/helpers/helpers.go
icarus-sparry: Looks good now. A bit long winded compared to C but that is go for you.

matthewrsj:
I've addressed the suggestions. Merging now with the approval.

PR #13, opened 2018-05-30
Bundle tools [DO NOT MERGE]

Gabib32:
Still need to add functions to get the includes, and packages lists. 
Also, going to add more test cases, and take out most of the clr-bundles definitions in the testdata


pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+	"regexp"
+	"strings"
+)
+
+var (
+	bundleHeaderFieldRegex = regexp.MustCompile(`^# \[([A-Z]+)\]:\s*(.*)$`)
+	includeBundleRegex     = regexp.MustCompile(`^include\(([A-Za-z0-9-_]+)\)$`)
+)
+
+type bundleHeader struct {
```
matthewrsj: Can just call this a `Header`, should be exported as well, especially if it contains exported fields.

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+	var b Definition
+	b.Includes = map[string]bool{"os-core": true}
+	b.DirectPackages = make(map[string]bool)
+	b.AllPackages = make(map[string]bool)
+
+	return b
+}
+
+// BundlesHash is a mapping of a bundle name (string) to the bundle definition
+// struct (DefStruct)
```
matthewrsj: s/DefStruct/Definition/

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+type BundlesHash map[string]*Definition
```
matthewrsj: I agree with @icarus-sparry that you don't need to define this type since it is only used in one place.

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+
+// GetPundleDef gets the pundle definition for a specified bundle. It sets then
+// header title as the package name, and gets all of the package dependents.
+// Specifically, the os-core packages, and itself.
+func GetPundleDef(name, bundlesDir string, b Definition) (*Definition, error) {
```
matthewrsj: Do we expect users to call this function? Since the difference between bundles and pundles are only in their definitions I think we don't need to expose that part of the implementation to the user.

If the user calls `bundle.GetBundle` on a pundle name they should get the same result as if they call `bundle.GetBundle` on a bundle name. The function should do the work of determining where the definition is.

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+	for inc := range parsed.Includes {
+		b.Includes[inc] = true
+	}
+	return nil
+}
+
+// GetDef gets the bundle definition for a specified bundle. This includes The
+// header information, directly depended packages, recursive bundle Includes
+// and recursive packages. It returns a bundle object.
+func GetDef(name, bundlesDir string, b Definition) (*Definition, error) {
```
matthewrsj: I don't like abbreviations in names. So I like `GetDefinition` over `GetDef`. I like names like `ReadDefinition` or `ParseDefinition` as well since that is what the function is doing.

matthewrsj: Here you are requiring the user pass in a `bundle.Definition` struct which you then populate and pass back a reference to. This is a strange way to do things because it requires the user to do something like this:
```go
b := bundle.NewDefinition()
bptr, err := bundle.GetDef("name", "path", b)
```

The flow that makes more sense to me is just letting `GetDefinition` do all the work:
```go
b, err := bundle.GetDefinition("name", "path") // b is a *bundle.Definition
```
The user doesn't need to use a constructor and then switch to using a pointer to the bundle they just created. I don't think the constructor is really necessary at all.

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+				b.AllPackages[line] = true
+			}
+		}
+	}
+	return &b, nil
+}
+
+// CheckDefType checks whether the item is a bundle or a pundle, then returns
+// the Definition with the appropriate information
+func CheckDefType(name, bundlesDir string) (*Definition, error) {
```
matthewrsj: The user doesn't need to know if it is a bundle or a pundle, so this doesn't need to be exposed.

Also, see my comment about not liking abbreviations.

pkg/bundle/bundle.go
```diff
@@ -0,0 +1,201 @@
...
+			return GetPundleDef(name, bundlesDir, NewDefinition())
+		}
+		return nil, fmt.Errorf("%s is neither a pundle or a bundle", name)
+	}
+	return GetDef(name, bundlesDir, NewDefinition())
+}
+
+// GetAll iterates the bundles/ directory, and calls the GetDef function to
+// create a Mapping of bundle name to bundle objects
+func GetAll(bundlesDir string) (BundlesHash, error) {
```
matthewrsj: just an opinion, but I prefer `ReadAll` since it gives a little better sense of what is happening. YMMV

pkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,100 @@
...
+// limitations under the License.
+
+package bundle
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/go-test/deep"
```
icarus-sparry: You need to alter the .travis.yml file to include this, since we are not using "go get" to set up the environment.

pkg/bundle/bundle.go
icarus-sparry: Repeating my suggestion from the branch, this function ought to get GetDef (or GetDefinition), and the existing GetDef ought to be renamed.

pkg/bundle/bundle.go
icarus-sparry: I comment below that this ought to be renamed, e.g. to GetBundleDefinition. Worth considering if it could be getBundleDefinition (i.e. not exported).

Gabib32: would bundle.g(G)etBundleDefinition be a no-no though with stuttering?

matthewrsj: Well as I mention elsewhere I think `ReadDefinition` or `ParseDefintion` is a more appropriate name... or maybe `LoadDefinition`

Gabib32:
Updated this code. Still need to add a GetDirectPackages and GetAllPackages function for a user to get a sorted list of package items. I figure they will be very similar to the GetIncludes function, so I am opting for a code review first, then I'll make the changes, add those functions and hopefully can rebase after that.

PR #15, opened 2018-06-04
Add Results definition and functions

matthewrsj:
Add a basic Results type that enables printing in different formats.
This new internal testutils package can be expanded to provide more
test-specific helper functions.


internal/testutils/results.go
```diff
@@ -0,0 +1,78 @@
...
+Skipped:  {{.Skipped}}
+
+{{range $f := .Failures}}
+{{.f}}
+{{end}}
+`
+
+// Print the Results to output if defined, otherwise print to stdout. Use the
+// resultTemplate.
+func (r *Results) Print(output string) error {
```
icarus-sparry: It would be better to pass an os.Writer to this, rather than a string to represent a filename to this function.

Of course you could still have a helper function which does the file open dance if you think this is wanted.

internal/testutils/results.go
```diff
@@ -0,0 +1,78 @@
...
+// PrintJSON prints the Results in JSON format to output if defined, otherwise
+// print to stdout.
+func (r *Results) PrintJSON(output string) error {
+	resOut, err := json.Marshal(r)
+	if err != nil {
+		return err
+	}
+
+	if output != "" {
+		return ioutil.WriteFile(output, resOut, 0655)
```
icarus-sparry: 0655 ? 0666 or 0644 would be more conventional, 0666 if you trust the user to set umask correctly, 0644 if you don't.

internal/testutils/results.go
```diff
@@ -0,0 +1,78 @@
...
+		f, err = os.OpenFile(output, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+		if err != nil {
+			return err
+		}
+	}
+
+	return t.Execute(f, r)
+}
+
+// PrintJSON prints the Results in JSON format to output if defined, otherwise
```
icarus-sparry: I have a couple of issues with this comment - output is always "defined", and I am trying to figure out the use case for this function.

As I commented above probably better to pass an os.Writer to this.

internal/testutils/results.go
```diff
@@ -0,0 +1,78 @@
...
+// Results holds the results of a test run
+type Results struct {
+	Passed   uint
+	Failed   uint
+	Skipped  uint
+	Total    uint
+	Failures []string
+}
+
+const resultTemplate = `
```
icarus-sparry: As you know I like using templates, but I am not sure that having a template adds anything here apart from overhead.

matthewrsj: I used a template so we could easily update the Results struct as needed and define the output. This is a naive first implementation as nothing needs it yet, so it is just a guess at how it should look.

internal/testutils/results.go
matthewrsj: Done, passing an io.Writer now.

internal/testutils/results.go
matthewrsj: Oops, changed to 0644

matthewrsj:
> Approved for when we get something that wants to use it.

See #16

PR #16, opened 2018-06-04
Add pipcheck subcommand and test

matthewrsj:
Add the 'diva pipcheck' subcommand. This iteration requires root access
to run, which actually breaks configuration parsing that tries to
determine the mixer workspace under $HOME. This issue exists as a TODO.

Only last commit is relevant, based on #15 

Edit: last *two* commits are relevant.


matthewrsj:
Note that this only prints to stdout in this current implementation.

cmd/pipcheck.go
```diff
@@ -0,0 +1,87 @@
...
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/clearlinux/diva/internal/config"
```
icarus-sparry: At some stage we are going to need to have a discussion about the use of "internal", but probably not now.

The point of internal is to allow packages that don't make sense on their own, and things like testutils is something that might get reused.

cmd/pipcheck.go
```diff
@@ -0,0 +1,87 @@
...
+	err = results.Print(os.Stdout)
+	if err != nil {
+		helpers.Fail(err)
+	}
+}
+
+// PipCheck runs 'pip check' in a chroot at path
+func PipCheck(path string) (*testutils.Results, error) {
+	r := &testutils.Results{}
+	_, err := helpers.RunCommandOutput("chroot", path, "/bin/bash", "-c", "pip check")
```
icarus-sparry: There should be a comment explaining why this is being run via bash, rather than directly.

internal/helpers/helpers.go
```diff
@@ -204,3 +204,9 @@ func RunCommandOutput(cmdname string, args ...string) (*bytes.Buffer, error) {
 	}
 	return &outBuf, nil
 }
+
+// Fail prints the error and exits the program with an error code
+func Fail(err error) {
+	fmt.Fprintf(os.Stderr, "ERROR: %s\n", err)
```
icarus-sparry: It is conventional in unix to print the program name with a failure, so if you have a pipeline 10 you stand some chance of guessing which thing objected.

Of course the counter to that is ed, which is (in)famous for its terseness - see   https://en.wikipedia.org/wiki/Ed_(text_editor)

cmd/pipcheck.go
matthewrsj: My thought was this would only be useful to Diva, but if you disagree I can make it a non-internal package. The reason I made a `testutils` package instead of throwing it all in `helpers` is because I didn't necessarily want to import `helpers` everywhere if it grows large.

PR #18, opened 2018-06-07
Bundle definition library

Gabib32:
A library that reads bundle/pundle information and creates
bundle definitions. A single bundle/bundle, or all bundles
within a clr-bundles directory can be constructed. The
definition has header information, include dependencies,
direct package dependencies, and all package dependencies.


bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+		pundle, err := GetDefinition(line, bundlesDir)
+		if err != nil {
+			return nil, err
+		}
+		bundles[line] = pundle
+	}
+	return bundles, nil
+}
+
+// GetIncludes returns a sorted list of all the includes for a specified bundle.
```
matthewrsj: `s/list/slice/`

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	bundle, err := GetDefinition(name, bundlesDir)
+	if err != nil {
+		return nil, err
+	}
+	var sorted []string
+	for p := range bundle.Includes {
+		sorted = append(sorted, p)
+	}
+	sort.Strings(sorted)
+	return sorted, nil
```
matthewrsj:
```go
return sort.Strings(sorted), nil
```

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// GetAllPackages returns a sorted list of all packages for a specified bundle.
+// AllPackages includes the direct packages for a bundle/pundle, along with the
+// direct packages of the bundle includes; so all package dependencies.
+func GetAllPackages(name, bundlesDir string) ([]string, error) {
```
matthewrsj: Maybe call this `GetAllPackagesForBundle` since I can envision a future where we might want to `GetAllPackages` referenced in `bundlesDir` as well.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	bundle, err := GetDefinition(name, bundlesDir)
+	if err != nil {
+		return nil, err
+	}
+	var sorted []string
+	for p := range bundle.AllPackages {
+		sorted = append(sorted, p)
+	}
+	sort.Strings(sorted)
+	return sorted, nil
```
matthewrsj:
```go
return sort.Strings(sorted), nil
```

bundle/bundle_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+func coreObject() Definition {
+	var core Definition
+
+	core.Header.Title = "os-core"
+	core.Includes = make(map[string]bool)
+	core.DirectPackages = make(map[string]bool)
+	core.AllPackages = make(map[string]bool)
+
+	oscorePackages := []string{"bash-bin", "ca-certs-static", "clr-power-tweaks",
```
matthewrsj: Instead of doing this, then iterating over it to create a map, just declare it all at once and deep copy one of the maps to the other.
```go
core := Definition{
    Header{
        Title: "os-core",
    },
    Includes: map[string]bool{},
    DirectPackages: map[string]bool{
        "bash-bin": true,
        ...,
    },
    AllPackages: make(map[string]bool),
}
for k := range core.DirectPackages {
    core.AllPackages[k] = true
}
return core
```

bundle/bundle_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	for _, p := range oscorePackages {
+		core.DirectPackages[p] = true
+		core.AllPackages[p] = true
+	}
+	return core
+}
+
+func TestHeaderScraper(t *testing.T) {
```
matthewrsj: The test should be named after the function being tested and the functionality it is testing, like `TestGetDefinitionValidHeader`

bundle/bundle_test.go
```diff
@@ -0,0 +1,263 @@
...
+				"",
+				"Developer Name <Developer@email.com>"},
+		},
+		{"joe",
+			Header{"joe", "", "", "", ""}},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.bundleName, func(t *testing.T) {
+			actual, err := GetDefinition(tc.bundleName, "testdata/clr-bundles")
```
matthewrsj: Instead of relying on data already on disk that makes it hard to understand this test since I can't easily see the data, you could be generating the data here and writing to a tempfile within a tempdir that is your clr-bundles repo. To understand if this test is correct I have to look for the bundle definition files under the testdata directory.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+func getPundleDefinition(name, bundlesDir string, pundle *Definition) (*Definition, error) {
+	pundles, err := ioutil.ReadFile(path.Join(bundlesDir, "packages"))
+	if err != nil {
+		return nil, err
+	}
+	if strings.Contains(string(pundles), "\n"+name+"\n") {
+		pundle.Header.Title = name
+		pundle.DirectPackages[name] = true
+		pundle.AllPackages[name] = true
+		return pundle, nil
```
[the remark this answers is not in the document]
Gabib32: That happens in the initialization of the Definition object.

bundle/bundle_test.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+}
+
+func (ti *testInstance) teardown(t *testing.T) {
+	if err := os.RemoveAll(ti.testdir); err != nil {
+	    t.Fatal(err)
+	}
+}
+
+func createBundleDefinitionTestFiles(bundleName, testdir string) error {
```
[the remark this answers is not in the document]
Gabib32: Would it be acceptable to call the addBundle method within the setup function to add os-core?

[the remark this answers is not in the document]
Gabib32: Also, instead of passing a different slice for each item; could I pass a map of string to slices; with the string being content type and the slice being the content. That way if I wanted to add a bundle with only header information, or only include information I could do that?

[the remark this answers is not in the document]
Gabib32:
```
content := map[string][]string{"header": []string{
		"# [TITLE]: os-core",
                "...",
	}, "packages": []string{"bash-bin",
		"ca-certs-static",
		"...",
	}}
```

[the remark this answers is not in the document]
matthewrsj: > Would it be acceptable to call the addBundle method within the setup function to add os-core?

Yes that seems reasonable.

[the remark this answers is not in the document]
matthewrsj: As for the other question, I don't think that makes sense for negative testing. I imagine a test where you want to put the header information *after* the body or with a ton of whitespace in-between the header and the body. Passing different strings for `header`, `includes`, and `packages` means you will always be writing the content in the same way.

If you just pass in all the content as a single string your negative testing can mess it all up intentionally and assert that a failure occurs.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	"path"
+	"path/filepath"
+	"reflect"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var (
+	bundleHeaderFieldRegex = regexp.MustCompile(`^# \[([A-Z]+)\]:\s*(.*)$`)
```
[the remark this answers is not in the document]
phmccarty: I'm okay with accepting both (with or without space). Eventually, clr-bundles is switching to TOML, right? After that time, DIVA can be more strict about stray space characters, since TOML strings will likely be interpreted verbatim.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	includeBundleRegex     = regexp.MustCompile(`^include\(([A-Za-z0-9-_]+)\)$`)
```
phmccarty: This is a nitpick, but I prefer the `-` to be either the first or last character in the regexp, to more easily see that it does not signify a character range. For example, either `^include\(([-A-Za-z0-9_]+)\)$` or `^include\(([A-Za-z0-9_-]+)\)$` instead.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+		if i == t {
+			return true
+		}
+	}
+	return false
+}
+
+func updateIncludes(packageInclude, bundlesDir string, b *Definition, visitedIncludes []string) error {
+	if inSlice(visitedIncludes, packageInclude) {
+		return errors.New("Recursive bundle includes")
```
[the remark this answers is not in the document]
phmccarty: I agree. Something like `"Bundle include loop detected"`.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	} // end reading file
+	return b, nil
+}
+
+func getPundleDefinition(name, bundlesDir string, pundle *Definition) (*Definition, error) {
+	pundles, err := ioutil.ReadFile(path.Join(bundlesDir, "packages"))
+	if err != nil {
+		return nil, err
+	}
+	if strings.Contains(string(pundles), "\n"+name+"\n") {
```
phmccarty: I'm not sure what's more idiomatic for Go, but perhaps we should iterate through the pundles string line-by-line instead, searching for a match? For example, by using `Scanner` from `bufio`?

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	Name   string
+	Header Header
+
+	Includes       map[string]bool
+	DirectPackages map[string]bool
+	AllPackages    map[string]bool
+}
+
+func initializeOsCore(bundlesDir string) error {
+	var err error
```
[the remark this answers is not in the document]
Gabib32: coreBundle is the global, and is already defined above, will the short order cause the coreBundle to only be defined within the scope of the initializedOsCore function?

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+		pundle.AllPackages[name] = true
+		return pundle, nil
+	}
+	return nil, fmt.Errorf("%s is neither a pundle or a bundle", name)
+}
+
+func getBundleDefinition(name, bundlesDir string, visitedIncludes []string) (*Definition, error) {
+	// check that the core bundle was initialized in the init() function. It must
+	// exist for proper bundle definition partion
+	if name != "os-core" && reflect.DeepEqual(coreBundle, &Definition{}) {
```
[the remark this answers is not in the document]
Gabib32: I don't want to continually recreate the os-core bundle with every getBundleDefinition call, so I was trying to store it globally so that the file didn't need to be read every time.

bundle/bundle.go
Gabib32: Error: sort.Strings(sorted) used as value

matthewrsj: Ahh, right my bad.

bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+func newDefinition() (Definition, error) {
+	var b Definition
+	b.Includes = make(map[string]bool)
+	b.DirectPackages = make(map[string]bool)
+	b.AllPackages = make(map[string]bool)
+
+	for include := range coreBundle.Includes {
+		b.Includes[include] = true
+	}
+	for pkg := range coreBundle.AllPackages {
```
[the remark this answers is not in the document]
Gabib32: They should be the same, but all of the os-core bundle packages should be in all of the other Definition.AllPackages, so that is what this is doing - copying the package.

bundle/bundle.go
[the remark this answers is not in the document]
matthewrsj: correct!

bundle/bundle.go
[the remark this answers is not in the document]
matthewrsj: :+1:

bundle/bundle.go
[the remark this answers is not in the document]
matthewrsj: ok

bundle/bundle_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+	// Create packages file within test directory
+	if _, err := os.Create(filepath.Join(testData.testdir, "packages")); err != nil {
+		testData.t.Fatal(err)
+	}
+
+	// Create os-core bundle definition test file
+	content := []string{"# [TITLE]: os-core",
+		"# [DESCRIPTION]: Run a minimal Linux userspace", "# [STATUS]: Active",
+		"# [CAPABILITIES]:", "# [MAINTAINER]: Super Smart Developer <email@email.com>",
```
matthewrsj: This must be me :rocket:

bundle/bundle_test.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// cleanup testdir and all of its content
+func (testData *testInstance) teardown() {
+	if err := os.RemoveAll(testData.testdir); err != nil {
+		testData.t.Error(err)
+	}
+}
+
+func (testData *testInstance) addBundle(name, dir string, content ...string) {
```
matthewrsj: Any reason you pass `content` as a `[]string` instead of just a `string`? You convert it back to a `string` below with `strings.Join`.

bundle/bundle_test.go
```diff
@@ -0,0 +1,355 @@
...
+	fulldir := filepath.Join(testData.testdir, dir)
+	if _, err := os.Stat(fulldir); err == nil && dir != "packages" {
+		if err := os.Remove(fulldir); err != nil {
+			testData.t.Fatal(err)
+		}
+	}
+
+	filemode := os.O_APPEND | os.O_CREATE | os.O_RDWR
+	fname, err := os.OpenFile(filepath.Join(testData.testdir, dir), filemode, 0644)
```
matthewrsj: You could achieve the same with 
```golang
err := ioutil.WriteFile(filepath.Join(testData.testdir, dir), strings.Join(content, "\n"), 0644)
```
and pretty much delete the next 10 lines.

bundle/bundle_test.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		testData.t.Fatal(err)
+	}
+
+	if _, err := fname.WriteString(strings.Join(content, "\n")); err != nil {
+		testData.t.Fatal(err)
+	}
+
+	if err := fname.Close(); err != nil {
```
matthewrsj: this is typically deferred right after opening,
```golang
f, err := os.OpenFile(...)
defer f.Close()
```
or
```golang
defer func() {
    _ = f.Close()
}
```
for linter reasons we've talked about.

bundle/bundle_test.go
```diff
@@ -0,0 +1,355 @@
...
+			[]string{"include(cycle3)"},
+		},
+	}
+
+	for _, bundle := range bundleAdds {
+		testData.addBundle(bundle.name, filepath.Join("bundles", bundle.name), bundle.content...)
+	}
+
+	_, err := GetDefinition("cycle2", testData.testdir)
+	if strings.TrimSpace(err.Error()) != "Bundle include loop detected" {
```
matthewrsj: You should make a custom test error here since the `err` will just say what the error is, not that you expected a certain error.
```golang
t.Fatalf("error %s did not match expected 'Bundle include loop detected'", err.Error())
```
You'd also wan't to make sure `err` isn't `nil` in this case.

bundle/bundle_test.go
icarus-sparry: Should use example.com, which is explicitly reserved for such use. email.com exists!

bundle/bundle_test.go
Gabib32: ioutil.Writefile takes a []byte slice not strings

matthewrsj: so just add `[]byte()` around the string

bundle/bundle_test.go
Gabib32: https://joeshaw.org/dont-defer-close-on-writable-files/ This was my reasoning for not using defer, but I'll make that switch.

bundle/bundle_test.go
Gabib32: Also, and more importantly, this doesn't allow packages to be appended to the file. It recreates a new file every time. If it is a package bundle, it needs to append to the packages file, instead of create a new file every time.

matthewrsj: ah, good point! never mind!

bundle/bundle_test.go
matthewrsj: That's a good argument, and I've used this method before
```golang
    defer func() {
        cerr := f.Close()
        if err == nil {
            err = cerr
        }
    }()
```

Gabib32: What is the err in this situation?

Gabib32: defer func() {
		if err := fname.Close(); err != nil {
			testData.t.Fatal(err)
		}
	}()

matthewrsj: > What is the err in this situation?

That was just an example. The `err` there comes from the outer scope so if `err` is nil it is set to `cerr` which could be non-nil as a result of the `Close` call. So that function will only override `err` if `err` is nil, otherwise `err` keeps its original value. Handy.

re: your code snippet: yeah that looks fine, probably doesn't need to be fatal though right?

icarus-sparry: This example also requires you to say something like ```return wibble, err``` everywhere, and have the same ```err``` variable in scope everywhere.

A different strategy is to avoid using defer to close the file, and instead do the close explicitly. Your task then is to make sure that all execution paths will get to this close. One way to do this is to put the stuff between the open and close into a function, so the code looks something like
````
func wibble()(){
    a, err := Open(...)
    if err != nil { ... }
    ret,err := func () {  //You essentially add this line
                  .... // original lines here
                  // can have returns in here
                } ()   // and add this lineNote the () so this is a call of a function literal.
    cerr := Close(a)  // have a temp place to hold the close error, we want to always call Close
    if err != nil { err = cerr } // use the first error which is not nil between err and cerr
    return ret, err
}
```

matthewrsj:
lgtm, @icarus-sparry anything else before I merge?

bundle/bundle.go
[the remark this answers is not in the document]
icarus-sparry: OK, let me ask this a different way. In what circumstances would you look at b.AllPackages rather than at a global coreBundle.AllPackages?

bundle/bundle.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// GetIncludesForBundle returns a sorted slice of all includes for a specified bundle.
+func GetIncludesForBundle(name, bundlesDir string) ([]string, error) {
+	bundle, err := GetDefinition(name, bundlesDir)
+	if err != nil {
+		return nil, err
+	}
+	var sorted []string
+	for p := range bundle.Includes {
```
icarus-sparry: Do we need a loop here, rather than just ```sorted = append(sorted, bundle.Includes)``` ((maybe we do to get the types correct?)

bundle/bundle.go
```diff
@@ -0,0 +1,274 @@
...
+// GetAllPackagesForBundle returns a sorted slice of all packages for a specified
+// bundle. AllPackages includes the direct packages for a bundle/pundle, along
+// with the direct packages of the bundle includes; so all package dependencies.
+func GetAllPackagesForBundle(name, bundlesDir string) ([]string, error) {
+	bundle, err := GetDefinition(name, bundlesDir)
+	if err != nil {
+		return nil, err
+	}
+	var sorted []string
+	for p := range bundle.AllPackages {
```
icarus-sparry: Ditto, do we need a loop?

bundle/bundle.go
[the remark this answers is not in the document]
matthewrsj: b.AllPackages will be a superset of coreBundle.AllPackages. Especially if A includes B includes C and all include os-core. A.AllPackages is relevant and different from coreBundle.AllPackages

bundle/bundle.go
matthewrsj: Oh right, you need to expand that arg though.
```golang
sorted = append(sorted, bundle.Includes...)
```

icarus-sparry: I knew that!

bundle/bundle.go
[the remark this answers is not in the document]
icarus-sparry: /me adjusts mental model, "AllPackages -> AllIncludedPackages"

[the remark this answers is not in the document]
Gabib32: I can change the naming convention if you guys think that will make it more logical / easier to read.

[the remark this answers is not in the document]
matthewrsj: Keep it this way for now since mixer uses that. We can revisit later.

bundle/bundle.go
Gabib32: bundle.Includes is a map[string]bool I'm not sure how to append only the keys without a for loop. Expanding the arg didn't work

matthewrsj: ahhhh, ok you're right.

PR #19, opened 2018-06-19
pkginfo: initial API implementation

matthewrsj:
This patch fills out the API with initial implementations of the
functions. Work still to be done is documented in TODO comments. This
patch allows users of the library to import RPMs from an RPM repository
and store them in a running redis database. Users can also retrieve all
repo information (RPMs and Files) using this API.

The getrepo.go functionality was moved to pkginfo because at least for
now it is the only consumer. It was extremely useful for getrepo
functionality to have usage of the Repo struct defined in types.go.


matthewrsj:
Initial API stubs for you guys to look at. LMK if anything looks off.

pkginfo/types.go
```diff
@@ -0,0 +1,58 @@
...
+// SRPMName is empty this indicates the RPM is already a source RPM. For binary
+// RPMs it will be populated with that RPMs associated source RPM name.
+type RPM struct {
+	Name         string
+	Version      string
+	Release      string
+	Architecture string
+	SRPMName     string
+	License      []string
+	Requires     []string
```
phmccarty: Do we want `BuildRequires []string` too, to be used by `GetBuildRequires()`?

matthewrsj: yeah probably.

pkginfo/types.go
```diff
@@ -0,0 +1,58 @@
...
+	Provides     []string
+	Files        []*File
+}
+
+// File contains all information for a file in an RPM.
+// Additional fields Name, Type, SwupdHash, and CurrentVersion are used by
+// swupd operations.
+type File struct {
+	Name           string
```
icarus-sparry: Not sure if we want to be using string in here, rather than our own datatype that interns strings. For ease of first implementation lets use string for the time being but be open to the idea that we might want to change it.

matthewrsj: How often will we have duplicated filenames though? Isn't it a failure case to have filename collisions outside the same SRPM?

icarus-sparry: I am thinking that when we are comparing two releases then almost every file will be duplicated.

matthewrsj: Ah, very good use-case. I agree.

matthewrsj:
Removing review requests, I realize now I haven't implemented the entire API yet.

matthewrsj:
Ok, adding you back. This is complete now.

matthewrsj:
Also, not sure of a good way to add unit tests for this, but open to suggestions.

pkginfo/download_repo.go
```diff
@@ -207,46 +207,73 @@ func downloadAllRPMs(packages []string, workingDir string) error {
...
 	}
 
-	packages, err := buildPackageURLs(flistsPath, version)
+	packages, err := buildPackageURLs(repo, flistsPath)
 	if err != nil {
-		return err
+		return "", err
+	}
+
+	return filepath.Join(workingDir, "packages"), downloadAllRPMs(packages, workingDir)
```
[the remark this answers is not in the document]
matthewrsj: `filepath.Join` for file paths. this has some more information: https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

pkginfo/redis.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// storeRepoInfoRedis stores all data in repo to the running redis-server
+func storeRepoInfoRedis(repo *Repo) error {
+	repoKey := fmt.Sprintf("%s%d%s", repo.Name, repo.Version, repo.Type)
+	if _, err := redisConn.Do("SET", repoKey, repo.URI); err != nil {
+		return err
+	}
+	defer func() {
+		_ = redisConn.Close()
```
Gabib32: What not print this error if not nil, so that there is knowledge of the redis connection not closing?

matthewrsj: So, I throw away the error here because there's nothing we can do about it at the point we are exiting the function. I don't want to print out the error because there is nothing the user can do about it either and an unactionable warning is just confusing.

icarus-sparry: Being realistic, the connection will be closed. What would potentially be more interesting is if the Close() can report any delayed errors.

In general printing out an error message from a library is not a great idea. You don't know how the library is going to be used. For example you might have a curses based front end, and printing something might corrupt the screen (and will almost certainly move the cursor).

So the choices are essentially to return an error, or set this as an error if there isn't an earlier one, or to "panic". If the latter is chosen then the fact that it can panic should be noted so the user of the library can use "recover" if desired.

Gabib32: Ahh okay, thank you both for the explanation!

matthewrsj:
Rebased

PR #20, opened 2018-06-23
Add fetch subcommand

matthewrsj:
The fetch subcommand allows users to fetch information to their system
for validation. Users can download and cache the latest bundles
definition files, an RPM repository at a specified version or latest,
and all update manifests at a specified version or latest.

~**based on #19** only the latest ~commit is~ two commits are relevant.~


cmd/fetch.go
```diff
@@ -0,0 +1,267 @@
...
+		u.url = allFlags.upstreamURL
+	}
+
+	if allFlags.version != 0 {
+		u.ver = allFlags.version
+		// no need to continue
+		return u, nil
+	}
+
+	resp, err := http.Get(u.url + "/latest")
```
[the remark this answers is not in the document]
matthewrsj: At this point diva *is* clear linux specific. The `pkginfo` package shouldn't be, but this is just the diva framework. Since we are operating on CL bundle definition file and CL update manifests keeping this part of the project CL-specific seems ok to me. If downstreams of Clear Linux are being used as the upstream, they should still be publishing all the files that official CL is publishing correct?

cmd/fetch.go
```diff
@@ -0,0 +1,267 @@
...
+	Run:   runFetchAllCmd,
+	Short: "Fetch metadata from <version> or latest available if <version> is not supplied",
+	Long: `Fetch latest bundle definitions, RPM repository, and update manifests from <version>.
+If <version> is not supplied, fetch latest available content. Pass
+--upstreamurl to fetch RPMs and update metadata from a location other than the
+configured Upstream URL and --bundleurl to fetch the bundle definitions from a
+location other than default.
+
+RPMs will be cached under the cache location defined in your configuration or
+default to $HOME/clearlinux/data/rpms/<version>. Bundle definition files will
```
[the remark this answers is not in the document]
matthewrsj: Are we allowing two different upstreams for validation? My assumption was one upstream plus however many local repos you want. The `pkginfo` library architecture reflects this assumption by putting upstream repos under the `clear` namespace.

cmd/fetch.go
```diff
@@ -0,0 +1,162 @@
...
+
+var allFlags allFetchFlags
+
+var fetchCmd = &cobra.Command{
+	Use:   "fetch",
+	Short: "Fetch and cache various build information",
+}
+
+var fetchAllCmd = &cobra.Command{
+	Use:   "all [--version <version>] [--bundleurl|--upstreamurl <url>]",
```
Gabib32: Can both --upstreamurl <url> and --bundleurl <url> be passed, or is it an xor?

diva/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return u, err
+	}
+
+	verString := strings.Trim(string(body), "\n")
+	v, err := strconv.ParseUint(verString, 10, 32)
+	u.Ver = uint(v)
+
+	// no support yet for modifying cache location via commandline
```
Gabib32: Maybe note to be careful here for when we do put support in, a version the function returns before it would get the cache location modification (line51).

diva/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+// GetLatestBundles clones or pulls the latest clr-bundles definitions to
+// conf.Paths.BundleDefsRepo
+func GetLatestBundles(conf *config.Config, url string) error {
+	if url == "" {
+		url = conf.BundleDefsURL
+	}
+
+	if _, err := os.Stat(conf.Paths.BundleDefsRepo); err == nil {
+		helpers.PrintBegin("pulling latest bundle definitions")
+		err = helpers.PullRepo(conf.Paths.BundleDefsRepo)
```
Gabib32: Why not just delete and clone every time? It feels safer, and keeps everything clean. Just a personal preference, take it or leave it :)

internal/config/config.go
```diff
@@ -42,9 +43,10 @@ type pathConfig struct {
 
 // Config struct that defines the layout of the configuration file
 type Config struct {
-	Mixer       mixConfig  `toml:"mixer"`
-	Paths       pathConfig `toml:"paths"`
-	UpstreamURL string     `toml:"upstream_url"`
+	Mixer         mixConfig  `toml:"mixer"`
+	Paths         pathConfig `toml:"paths"`
+	UpstreamURL   string     `toml:"upstream_url"`
+	BundleDefsURL string     `toml:"bundles_url"`
```
Gabib32: Isn't there a BundleDefsUrl in Paths/bundlesDefsRepo?

diva/fetch.go
```diff
@@ -0,0 +1,139 @@
...
+	u.Ver = uint(v)
+
+	// no support yet for modifying cache location via commandline
+	u.CacheLoc = conf.Paths.CacheLocation
+	return u, err
+}
+
+// FetchRepo fetches the RPM repo at the u.URL baseurl to the local cache
+// location
+func FetchRepo(u UInfo) error {
```
Gabib32: Could we consolidate this function with pkginfo.GetUpstreamRepoFiles? There seems to be some duplication?

internal/config/config.go
matthewrsj: err... yes...

matthewrsj: oh, actually no. That is the path on the system to where that repository is stored.

diva/fetch.go
matthewrsj: Just because it is a git repo, and by default I don't to throw away people's local branches and work.

diva/fetch.go
matthewrsj: Right now the setup is slightly different and GetUpstreamRepoFiles takes only a version as the argument and assumes a "clear" upstream. Ideally in the future the FetchRepo function will be much more versatile than it is now and will look different.

matthewrsj:
Merging now based on @Gabib32's approval. We can make improvements as needed. #31 is based on this PR.

PR #26, opened 2018-06-27
Bundle verify command implementation 

Gabib32:
A first attempt at bundle-verify, still need to add package deletes, and passing in one bundle, but wanted to be sure I was on the right track first.


cmd/bundle-verify.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// flags passed in as args
+var bundleFlags bundleCmdFlags
+
+// config object used by GetUpstreamRepoFiles and called functions
+var c *config.Config
+
+var verifyBundlesCmd = &cobra.Command{
+	Use:   "verify-bundles",
```
matthewrsj: Since this is a subcommand of `check` let's just call this `bundles` since verify and check are pretty much synonyms.
```
diva check bundles
```

cmd/bundle-verify.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func runVerifyBundle(cmd *cobra.Command, args []string) {
+	var err error
+	c, err = config.ReadConfig("")
+	if err != nil {
+		helpers.Fail(err)
+	}
+
+	if bundleFlags.version == 0 && bundleFlags.repoPath == c.UpstreamURL {
```
matthewrsj: noice

cmd/bundle-verify.go
```diff
@@ -0,0 +1,128 @@
...
+	Short: "Verify named packages exist in a repo",
+	Long: `Verify bundles are complete by checking that all named packages within
+bundle and package bundle files can be found in the configured repo. A <version>
+or <path> must be passed, and will run against all bundles or a <bundle>/`,
+	Run: runVerifyBundle,
+}
+
+func runVerifyBundle(cmd *cobra.Command, args []string) {
+	var err error
+	c, err = config.ReadConfig("")
```
matthewrsj: check out #20, once that is merged you can access the global `conf` variable instead of reading it in for every subcommand. No AR right now.

cmd/bundle-verify.go
```diff
@@ -0,0 +1,128 @@
...
+		helpers.Fail(err)
+	}
+
+	if bundleFlags.version == 0 && bundleFlags.repoPath == c.UpstreamURL {
+		helpers.Fail(errors.New("must supply either --version or --path argument"))
+	}
+
+	repo := pkginfo.Repo{
+		URI:  bundleFlags.repoPath,
+		Name: "mash",
```
matthewrsj: This should be something passed in, not hard-coded. Someone may want to verify upstream repo from the upstream URL.

cmd/bundle-verify.go
```diff
@@ -0,0 +1,128 @@
...
+		helpers.Fail(err)
+	}
+
+	if results.Failed > 0 {
+		os.Exit(1)
+	}
+}
+
+// BundleCompleteCheck checks the bundle packages are a repo
+func BundleCompleteCheck(repo *pkginfo.Repo, bundles bundle.Set) (*testutils.Results, error) {
```
matthewrsj: Naming nit: I like functions like this to be a verb, so to me this would read better as `CheckBundleComplete`.

Is there a reason this function is exported?

Gabib32: I was using the PipCheck as a template *shrugs* :)

matthewrsj: oof you win, but PipCheck is named so because it literally runs `pip check`

Gabib32: Oh I was referring more to the exporting of it - changing the name works for me :)

cmd/bundle-verify.go
```diff
@@ -0,0 +1,148 @@
...
+	if err := pkginfo.ImportAllRPMs(&repo); err != nil {
+		helpers.Fail(err)
+	}
+
+	bundles := make(bundle.Set)
+
+	if bundleFlags.bundle != "" {
+		singleBundle, err := bundle.GetDefinition(bundleFlags.bundle, c.Paths.BundleDefsRepo)
+		if err != nil {
+			helpers.Fail(err)
```
matthewrsj: Will this just `os.Exit(1)` without adding a failure to the test result if an include loop is encountered?

cmd/bundle-verify.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		bundles[singleBundle.Name] = singleBundle
+	} else {
+		var err error
+		bundles, err = bundle.GetAll(c.Paths.BundleDefsRepo)
+		if err != nil {
+			helpers.Fail(err)
```
matthewrsj: same question about the include loop

cmd/bundle-verify.go
```diff
@@ -0,0 +1,148 @@
...
+		helpers.Fail(err)
+	}
+
+	if results.Failed > 0 {
+		os.Exit(1)
+	}
+}
+
+// CheckBundleComplete checks the bundle packages are in the specified repo
+func CheckBundleComplete(repo *pkginfo.Repo, bundles bundle.Set) (*testutils.Results, error) {
```
matthewrsj: might as well not return an error at all if we are just going to throw it away.

pkginfo/rpm.go
```diff
@@ -44,6 +46,7 @@ func GetRPM(repo *Repo, rpm string) (*RPM, error) {
 	defer func() {
 		_ = c.Close()
 	}()
+	fmt.Println("EHERERERERER")
```
matthewrsj: :+1: :100: :laughing: :rofl: :robot: :rocket: :rocket: :1st_place_medal:

pkginfo/import.go
```diff
@@ -97,12 +97,12 @@ func fileFromPackageFile(pkgFI *rpm.FileInfo) *File {
 		Name:           pkgFI.Name(),
 		Type:           t,
 		Size:           uint(pkgFI.Size()),
-		Hash:           pkgFI.Digest(),
+		Hash:           "", // TODO
```
matthewrsj: no no no

internal/helpers/helpers.go
```diff
@@ -134,7 +134,7 @@ func xzExtractURL(url, target string) error {
 		return err
 	}
 
-	return RunCommandSilent("unxz", "-T", "0", target+".xz")
+	return RunCommandSilent("unxz", "-f", "-T", "0", target+".xz")
```
matthewrsj: +1 to this change but it should be in its own commit.

pkginfo/download_repo.go
```diff
@@ -38,12 +38,9 @@ func buildFilelistsURL(repo *Repo, workingDir string) (string, error) {
 	// download repomd.xml
 	repomdFile := filepath.Join(workingDir, "repomd.xml")
 	repomdURL := fmt.Sprintf("%s/repodata/repomd.xml", repo.URI)
-	_, err := os.Stat(repomdFile)
+	err := helpers.Download(repomdURL, repomdFile)
```
matthewrsj: +1 to this change but it should be in its own commit.

pkginfo/import.go
```diff
@@ -97,12 +97,12 @@ func fileFromPackageFile(pkgFI *rpm.FileInfo) *File {
...
 		SwupdHash:      "", // TODO
-		Permissions:    pkgFI.Mode().Perm().String(),
+		Permissions:    "", // TODO
```
matthewrsj: no no no

pkginfo/import.go
```diff
@@ -97,12 +97,12 @@ func fileFromPackageFile(pkgFI *rpm.FileInfo) *File {
...
 		Owner:          pkgFI.Owner(),
 		Group:          pkgFI.Group(),
-		SymlinkTarget:  pkgFI.Linkname(),
+		SymlinkTarget:  "", // TODO
```
matthewrsj: no no no

pkginfo/rpm.go
matthewrsj: ermahgerd

matthewrsj: :selfie:

matthewrsj: :cow: :cow2: :cowboy_hat_face:

pkginfo/import.go
Gabib32: That is a bad rebase - my b

cmd/bundle-verify.go
Gabib32: Yeah, do you want it to be added to the test suite? Should all errors?

matthewrsj: This at least should be added to the test suite I think.

Gabib32: Do we want the same logic for the importing of RPMS?

matthewrsj: Eh, maybe not for that. Can't trust the validation if we don't get all the RPMs

cmd/bundle-verify.go
```diff
@@ -0,0 +1,183 @@
...
+	if bundleFlags.version == 0 {
+		repo.URI = fmt.Sprintf("%s/mash/clear/x86_64/os/", bundleFlags.repoPath)
+	}
+
+	if err := pkginfo.ImportAllRPMs(&repo); err != nil {
+		helpers.Fail(err)
+	}
+
+	result := &diva.Results{Name: "bundle-verify"}
+	result, bundles := checkAndGetBundleDefinitions(result)
```
matthewrsj: you don't need to actually return `result` from these functions, you can just modify them in the function itself.

Gabib32:
Resolves issue #39

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+
+type bundleCmdFlags struct {
+	repoPath string
+	repoName string
+	version  uint
+	bundle   string
+}
+
+// flags passed in as args
+var bundleFlags bundleCmdFlags
```
phmccarty: Can you move this definition above the `init()` function? At first I couldn't immediately tell where the `bundleFlags` variable came from in `init()` until I read further...

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+
+// config object used by GetUpstreamRepoFiles and called functions
+var c *config.Config
```
phmccarty: Same comment as above.

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+bundle and package bundle files can be found in the configured repo. A <version>
+or <path> must be passed, and will run against all bundles or a <bundle>. An
+optional <reponame> argument may also be passed; the default being "clear". This
+check also ensures no include loops exist, and that the bundle filename matches
+the bundle definition header title.`,
+	Run: runVerifyBundle,
+}
+
+func runVerifyBundle(cmd *cobra.Command, args []string) {
+
```
phmccarty: remove blank line?

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+	repo := pkginfo.Repo{
+		URI:     fmt.Sprintf("%s/releases/%d/clear/x86_64/os/", bundleFlags.repoPath, bundleFlags.version),
+		Name:    bundleFlags.repoName,
+		Version: bundleFlags.version,
+		Type:    "B",
+	}
+
+	if bundleFlags.version == 0 {
+		repo.URI = fmt.Sprintf("%s/mash/clear/x86_64/os/", bundleFlags.repoPath)
```
phmccarty: Is this the default URI we want for the `version == 0` case? There are no repos on download.clearlinux.org that have "mash" in the name, so I'm wondering how this case should be handled... Maybe use "current"?

For the mash case, I would prefer having a separate option (or options) to validate it instead.

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+	return bundles
+}
+
+func checkBundleHeaderTitleMatchesFile(bundles bundle.Set, result *diva.Results) *diva.Results {
+	name := "bundle filename and title matching"
+	desc := "Ensure that the bundle header 'TITLE' and bundle file name are equal."
+
+	for filename, bundle := range bundles {
+		if filename != bundle.Header.Title {
+			result.Add(name, desc, fmt.Errorf("Bundle filename '%s' does not match Header title '%s'", filename, bundle.Header.Title), false)
```
phmccarty: Since this line is getting long, maybe store the formatted error string in a variable and then reference it on this line?

cmd/bundle-verify.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func checkIfPundleDeletesExist(result *diva.Results) *diva.Results {
+	name := "package bundle delete"
+	desc := "Determine whether a package bundle was deleted"
+
+	var deleted []string
+	var err error
+
+	output, err := helpers.RunCommandOutput("git", "-C", c.Paths.BundleDefsRepo, "diff", "latest...HEAD", "packages")
```
phmccarty: I would use `latest..HEAD` instead. The triple-dot syntax is useful for different situations where merge commits are involved or if commits are being compared from different branches.

cmd/bundle-verify.go
matthewrsj: Yeah this should be configurable instead of hardcoding "mash" in here. We should probably make that whole path after the initial %s part of a configuration option like "repo" or something. Maybe it would be easier to require it be passed as a flag?

Gabib32: I'm trying to think of the best way to solve this, because the only difference between the repo path of a version and of a mash version is the "/releases/" portion. Other than that, its: <url>/releases/<ver>/<mash_repo>/x86_64/os, which is nice because mash is just the VER. If we create the repopath variable in the config, would the default for version be missing, or 0, or some other placeholder?

Gabib32: I feel like if we were passing some sort of flag, we defeat the purpose of changing version to be a string?

matthewrsj:
The first commit has already been merged in master and is causing the merge conflicts. Please drop it.

PR #29, opened 2018-06-29
Modify pipcheck to display failure results

Gabib32:
Instead of failing if the pipcheck fails, display it as the
result output. Also modify the results display to provide a
seperator between the error code, and the message.


cmd/pipcheck.go
```diff
@@ -78,11 +75,11 @@ func runPipCheck(cmd *cobra.Command, args []string) {
...
 // PipCheck runs 'pip check' in a chroot at path
-func PipCheck(path string) (*testutils.Results, error) {
+func PipCheck(path string) *testutils.Results {
 	r := &testutils.Results{Name: "pipcheck"}
 	name := "pipcheck"
 	desc := "run pip check in full build root to check for missing python requirements"
-	_, err := helpers.RunCommandOutput("chroot", path, "pip", "check")
-	r.Add(name, desc, err, false)
-	return r, err
+	_, output := helpers.RunCommandOutput("chroot", path, "pip", "check")
```
matthewrsj: this is still the error return from `RunCommandOutput` so it should be called `err`.

We are throwing away the output, so we should actually invoke `RunCommandSilent` here.

matthewrsj:
Rebase this to master to drop the `V0 do not merge` commit.

PR #30, opened 2018-06-29
Modify bundles to remove trailing whitespace

Gabib32:
When a trailing space was at the end of a bundle include, it
was not matching regex matching for includes, so the line was
being considered a direct package for a bundle. To solve the
problem, trim whitespace for each line being read before
performing pattern matching. Include associated unit test for
the bug.


bundle/bundle_test.go
```diff
@@ -306,6 +306,75 @@ func TestCorrectEntireBundleDefinitionKoji(t *testing.T) {
...
+		content []string
+	}{
+		{name,
+			[]string{"# [TITLE]: package-utils",
+				"# [DESCRIPTION]: Utilities for packages",
+				"# [STATUS]: WIP",
+				"", // empty string
+				"# [CAPABILITIES]:",
+				"include(python3-basic)",
+				"# sad",                   // Out of order comment
```
matthewrsj: haha

PR #31, opened 2018-06-29
Add updatecontent checks and update results API

matthewrsj:
Add checks to validate update content. Verify that the published files
match the hashes listed in the manifests and the manifest files match
the hashes listed in the MoM. Verify that the zero packs contain all
necessary full files for each bundle and the full files match the hashes
in the manifest.

helpers: preserve permissions when untarring

Also update the results API to use go-tap protocol.


matthewrsj:
Question: I do not cache zero packs and instead extract them to a tempdir. I *do* cache the fullfiles, should I instead just throw them in a tempdir as well? It really speeds up the run if you have the files around but a) they take up a lot of space and b) how common will it be for people to re-run this check on the same content?

matthewrsj:
ping

cmd/fetch.go
```diff
@@ -150,7 +153,8 @@ func runFetchRepoCmd(cmd *cobra.Command, args []string) {
 }
 
 func runFetchUpdateCmd(cmd *cobra.Command, args []string) {
-	u, err := diva.GetUpstreamInfo(conf, allFlags.upstreamURL, allFlags.version)
+	fmt.Println(allFlags.recursive)
```
Gabib32: debug statement? ::cowboyhatmanface:: HAHHA

cmd/root.go
```diff
@@ -59,10 +59,12 @@ var conf *config.Config
 
 func initConfig() {
 	var err error
+	fmt.Println(rootCmdFlags.configPath)
```
Gabib32: ???

cmd/root.go
```diff
@@ -59,10 +59,12 @@ var conf *config.Config
...
 	conf, err = config.ReadConfig(rootCmdFlags.configPath)
 	if err != nil {
 		helpers.Fail(err)
 	}
+	fmt.Println(conf)
```
Gabib32: tsk tsk tsk

matthewrsj: :elephant: (tusk tusk tusk)

matthewrsj:
@bryteise requesting review from you since you wrote a nifty tool called swupdbb that this is emulating.

cmd/updatecontent.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/clearlinux/diva/diva"
+	"github.com/clearlinux/diva/internal/helpers"
+	"github.com/clearlinux/diva/updatecontent"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	checkCmd.AddCommand(ucCmd)
+	ucCmd.Flags().UintVarP(&ucFlags.version, "version", "v", 0, "version to check")
```
Gabib32: Will need to change with the int -> string implementation.

cmd/updatecontent.go
```diff
@@ -0,0 +1,120 @@
...
+	v := ucFlags.version
+	if v == 0 {
+		v, err = helpers.GetLatestVersion(conf.UpstreamURL)
+		if err != nil {
+			helpers.Fail(err)
+		}
+	}
+
+	results, err := UCCheck(v, ucFlags.recursive)
+	if err != nil {
```
Gabib32: Do we want the errors to be failures, or added to the testsuite results?

internal/helpers/helpers.go
```diff
@@ -225,3 +244,22 @@ func Fail(err error) {
 	fmt.Fprintf(os.Stderr, "%s: ERROR: %s\n", os.Args[0], err)
 	os.Exit(1)
 }
+
+// GetLatestVersion returns the version value at upstreamURL/latest or an error
+// if unable to do so.
+func GetLatestVersion(upstreamURL string) (uint, error) {
```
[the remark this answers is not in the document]
Gabib32: Also, latest could be found by using "current" as the version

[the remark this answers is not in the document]
matthewrsj: What's the URL though? https://download.clearlinux.org/current/ is a directory and https://download.clearlinux.org/update/current/ doesn't exist. I need the get the string of the current version.

cmd/updatecontent.go
matthewrsj: The intention here was to return an error when we couldn't even get the tests to run (the fetch fails).

diva/fetch.go
```diff
@@ -128,13 +141,127 @@ func FetchUpdate(u UInfo) error {
...
+
+				if f.err != nil {
+					fmt.Println(f.err)
+					if !retried {
+						fmt.Println("retrying download...")
+						retried = true
+						i--
+					} else {
+						fmt.Println("unable to download, continuing...")
+						errChan <- f.err
```
tmarcu: This will block forever if you hit more than 1 error because you have no reader to pull errors off the channel; just change errChan to be a buffered channel and you can leave the rest of the code the same :smiley:

updatecontent/check.go
```diff
@@ -0,0 +1,427 @@
...
+			}
+		}()
+	}
+
+	for _, f := range m.Files {
+		fCh <- f
+	}
+	close(fCh)
+	wg.Wait()
+	close(eCh)
```
tmarcu: The receiver should never close the channel IMO, only the sender, and if you use a buffered channel you can remove this.

updatecontent/check.go
```diff
@@ -0,0 +1,427 @@
...
+		}()
+	}
+
+	for v := range vers {
+		vCh <- v
+	}
+
+	close(vCh)
+	wg.Wait()
+	close(errCh)
```
tmarcu: Same comment as the other sections, I don't think the receiver should close this, and it should be buffered so routines can freely write to it as they hit errors.

updatecontent/check.go
```diff
@@ -0,0 +1,427 @@
...
+			}
+		}()
+	}
+
+	for _, b := range MoM.Files {
+		bCh <- b
+	}
+	close(bCh)
+	wg.Wait()
+	close(eCh)
```
tmarcu: remove

updatecontent/check.go
```diff
@@ -0,0 +1,427 @@
...
+			}
+		}()
+	}
+
+	for _, f := range m.Files {
+		fCh <- f
+	}
+	close(fCh)
+	wg.Wait()
+	close(eCh)
```
tmarcu: remove

diva/fetch.go
```diff
@@ -66,6 +69,16 @@ func GetUpstreamInfo(conf *config.Config, upstreamURL string, version string) (U
 	u.Ver = strings.Trim(string(body), "\n")
 	// no support yet for modifying cache location via commandline
 	u.CacheLoc = conf.Paths.CacheLocation
+
+	if !recursive {
+		var mv int
+		mv, err = strconv.Atoi(u.Ver)
```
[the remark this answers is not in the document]
matthewrsj: Nope, ParseUint returns a uint64, not a uint.

internal/helpers/helpers.go
```diff
@@ -73,7 +75,18 @@ func Download(url, filename string) error {
 	}
 
 	// move tempfile to final now that everything else has succeeded
-	return os.Rename(tmpFile, filename)
+	return renameIfNE(tmpFile, filename)
+}
+
+func renameIfNE(src, dst string) error {
```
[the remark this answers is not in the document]
matthewrsj: renameIfNotExists, I'll change the name

cmd/updatecontent.go
```diff
@@ -0,0 +1,112 @@
...
+	recursive bool
+}
+
+var ucFlags ucCmdFlags
+
+var ucCmd = &cobra.Command{
+	Use:   "updatecontent",
+	Short: "Validate update file and pack content",
+	Long: `Validate update content for <version> or latest if --version was not provided.
+Validates that all file and pack content is available and correct and the m
```
Gabib32: m? typo.

Gabib32: Also "Validates that all file and pack content is available and correct and the local configuration and <version>." doesn't really make much sense.

matthewrsj: Yikes lol.

cmd/updatecontent.go
```diff
@@ -0,0 +1,112 @@
...
+on all update content reachable through the manifests, otherwise validate only
+the current version.`,
+	Run: runUCCheck,
+}
+
+func runUCCheck(cmd *cobra.Command, args []string) {
+	var err error
+	v := ucFlags.version
+	if v == 0 {
+		v, err = helpers.GetLatestVersion(conf.UpstreamURL)
```
Gabib32: +1 making this a helper function!

cmd/updatecontent.go
matthewrsj: fixed

diva/fetch.go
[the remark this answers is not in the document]
tmarcu: Change all your code then ;P

updatecontent/check.go
```diff
@@ -0,0 +1,490 @@
...
+
+	var failures []string
+	es := checkBundleFileHashesPack(filepath.Join(tmpDir, "staged"), m, 0)
+	for _, e := range es {
+		failures = append(failures, e.Error())
+	}
+	return failures, nil
+}
+
+func checkSingleDelta(deltaFile, fromFile, expHash string) error {
```
Gabib32: I thought we weren't checking the deltas, only the delta packs?

PR #47, opened 2018-07-03
Change version variable to be string type

Gabib32:
Set version to be a string instead of a uint, so that "mash"
can be passed in as a version.


diva/fetch.go
```diff
@@ -31,21 +30,21 @@ import (
...
+func GetUpstreamInfo(conf *config.Config, upstreamURL string, version string) (UInfo, error) {
 	u := UInfo{}
 	if upstreamURL == "" {
 		u.URL = conf.UpstreamURL
 	} else {
 		u.URL = upstreamURL
 	}
 
-	if version != 0 {
+	if version != "0" || version == "mash" {
```
icarus-sparry: Logic looks wrong. If the version is "mock" then version != "0" (as it is "mock") so the second half of the "OR" is not needed. I suspect you want  ```if version != "0" && version != "mock" {``` or else expand it as a switch statement.

diva/fetch.go
```diff
@@ -63,10 +62,8 @@ func GetUpstreamInfo(conf *config.Config, upstreamURL string, version uint) (UIn
 		return u, err
 	}
 
-	verString := strings.Trim(string(body), "\n")
-	v, err := strconv.ParseUint(verString, 10, 32)
-	u.Ver = uint(v)
-
+	u.Ver = strings.Trim(string(body), "\n")
+	fmt.Println("UVER: ", u.Ver)
```
matthewrsj: debug 🖨

diva/fetch.go
matthewrsj: This was just checking the zero value of the uint version before you changed it. If the user manually sets a version the version detection should not continue, regardless of the contents of the variable. This check should not look like `if version != ""` as `""` is the zero-value of the string.

matthewrsj: Also I don't want to hard-code "mash" anywhere if we can help it. Making `version` a string should prevent us from having to do that since the user can name any arbitrary version they want.

matthewrsj:
Oh shoot, need to rebase

Gabib32:
rebased

PR #48, opened 2018-07-10
Vendor required dependencies

tmarcu:


icarus-sparry:
Are you going to change the travis file to no longer load these vendored packages?

matthewrsj:
So, I'm not exactly sure why this is failing, but you vendored `redigomock` when you shouldn't have since it is a test dependency.

matthewrsj:
@tmarcu are you going to debug this or do we want to wait to see what go1.11/go modules look like?

matthewrsj:
https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring

tmarcu:
Maybe we should wait and see what the new options will be. I don't really want to debug that & if it's getting obsoleted might as well go with the newest.

matthewrsj:
Right. Closing.

PR #54, opened 2018-07-17
Build bloat check

tmarcu:
Throwing this out to get a feel if this is on the right track, let me know if anything stands out immediately.


bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,134 @@
...
+	includes, err := bundle.GetIncludesForBundle(m.Name, "/home/tmarcu/clearlinux/projects/clr-bundles")
+	if err != nil {
+		fmt.Printf("Failed to get includes for manifest %s\n", m.Name)
+		return err
+	}
+
+	for _, i := range includes {
+		manifestFile := findManifest(i, mom.Files)
+		if manifestFile == nil {
+			fmt.Println("Couldn't find " + i)
```
tmarcu: There are still some debug prints like this in here that I will remove.

bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,134 @@
+package bloatcheck
```
matthewrsj: Needs license header

bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	includes, err := bundle.GetIncludesForBundle(m.Name, "/home/tmarcu/clearlinux/projects/clr-bundles")
+	if err != nil {
+		fmt.Printf("Failed to get includes for manifest %s\n", m.Name)
+		return err
+	}
+
+	for _, i := range includes {
+		manifestFile := findManifest(i, mom.Files)
```
matthewrsj: Sort mom.Files once then do a `sort.Search` in `findManifest` every time. Right now this is O(n^2)

bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		if manifestFile == nil {
+			fmt.Println("Couldn't find " + i)
+			return err
+		}
+		baseCache := filepath.Join(u.CacheLoc, "update")
```
matthewrsj: declare `baseCache` before the loop since it isn't changing.

bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,134 @@
...
+func GetBundleSize(u diva.UInfo) (map[string]int64, error) {
+	manifests, err := getManifests(u)
+
+	var wg sync.WaitGroup
+	bundleWorkers := len(manifests)
+	mChan := make(chan *swupd.Manifest)
+	errChan := make(chan error, bundleWorkers)
+	wg.Add(bundleWorkers)
+
+	var bundleSizes = make(map[string]int64)
```
matthewrsj: Make it a uint64 for that sweet sweet extra bit?

cmd/check.go
```diff
@@ -14,14 +14,128 @@
...
+	"github.com/clearlinux/diva/bloatcheck"
+	"github.com/clearlinux/diva/diva"
+	"github.com/clearlinux/diva/internal/helpers"
+	"github.com/spf13/cobra"
+)
+
+var highPrioBundles = map[string]bool{"os-core": true, "os-core-update": true, "c-basic": true, "kernel": true}
+
+type bloatCheckCmdFlags struct {
+	print bool
```
matthewrsj: Rename to something that isn't a go builtin https://golang.org/pkg/builtin/#print

cmd/check.go
```diff
@@ -14,14 +14,128 @@
...
 	Short: "Run various content and metadata checks",
 	Long:  `Run various checks against distribution content or metadata`,
 }
 
+func checkSize(name *string, sizeDiff, size float64) bool {
+	if _, ok := highPrioBundles[*name]; ok {
+		// High priority bundles cannot increase by more than 10% because they
+		// may affect many other bundles and minimal installations -> error out
+		if sizeDiff > (size * .1) {
+			helpers.Fail(fmt.Errorf("%s has increased by more than 10%%", *name))
```
matthewrsj: Should the 10% and 20% below be configurable?

cmd/check.go
```diff
@@ -14,14 +14,128 @@
...
+
+	if len(args) == 1 {
+		fmt.Printf("Size information for build %v\n", u.Ver)
+		for bundle, size := range fromBundleSizes {
+			fmt.Printf("%s: %d\n", bundle, size)
+		}
+	}
+
+	// Only do build compare if we have two versions passed in
+	if len(args) > 1 {
```
matthewrsj:
```go
if len(args) != 2 {
    // failure message
}
// do everything else
```

saves you some indentation yo

bloatcheck/bloatcheck.go
tmarcu: Overflow wraparound if I do that depending on which direction size changes :(

cmd/check.go
tmarcu: Yeah totally, should I leave those as defaults but override if cmd line given? I'll set the 10 & 20 as default in cobra if no args passed.

cmd/check.go
tmarcu: ty

bloatcheck/bloatcheck.go
matthewrsj: but that *sweet sweet* extra bit...

lol jk +1

cmd/check.go
matthewrsj: yeah I like that. This seems too specific for our config file.

cmd/check.go
tmarcu: wait no that won't work, passing in 1 value is OK if you just want bundle sizes for 1 version, if we check != 2 it'll always fail in that case.

tmarcu:
@matthewrsj @Gabib32 I only pushed a "Fix Comments" commit so it's easier to review, I'll squash it in before merging if you think it looks good.

bloatcheck/bloatcheck.go
matthewrsj: Time to remove them?

internal/helpers/helpers.go
```diff
@@ -27,6 +27,22 @@ import (
...
 
+// Min returns the max of two integers...since go only implements Float64
+func Min(a, b string) string {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+// Max returns the max of two integers...since go only implements Float64
```
matthewrsj: Max of two strings.

internal/helpers/helpers.go
```diff
@@ -27,6 +27,22 @@ import (
 	"strings"
 )
 
+// Min returns the max of two integers...since go only implements Float64
```
matthewrsj: Min of two strings

matthewrsj:
@Gabib32 take a look, feel free to merge when ready.

matthewrsj:
Actually, squash before merge

bloatcheck/bloatcheck.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+// Concurrently gets the size for each manifest
+func getSizes(u diva.UInfo, m *swupd.Manifest, mom *swupd.Manifest, bundleSizes map[string]int64) error {
+	if m.Name == "os-core-update-index" {
+		return nil
+	}
+
+	includes, err := bundle.GetIncludesForBundle(m.Name, "/home/tmarcu/clearlinux/projects/clr-bundles")
```
Gabib32: conf.Paths.BundlePath?

tmarcu: BundleDefsRepo right?

PR #55, opened 2018-07-17
Provide option to update RPMs

Gabib32:
If an update flag is passed, check whether an RPM already
exists for that package, and if so remove the cached
RPM and download the new one.


cmd/fetch.go
```diff
@@ -114,7 +117,7 @@ func runFetchAllCmd(cmd *cobra.Command, args []string) {
 		helpers.Fail(err)
 	}
 
-	err = diva.FetchRepo(u)
```
matthewrsj: Does it make sense to add `update` to the `UInfo` struct so we don't have to pass in this extra boolean everywhere?

matthewrsj:
commit: Add force option to extraction

Please put *why* you added the force option to extraction in the commit message.

pkginfo/download_repo.go
```diff
@@ -34,13 +34,13 @@ var c *config.Config
...
-func buildFilelistsURL(repo *Repo, workingDir string) (string, error) {
+func buildFilelistsURL(repo *Repo, workingDir string, update bool) (string, error) {
 	// download repomd.xml
 	repomdFile := filepath.Join(workingDir, "repomd.xml")
 	repomdURL := fmt.Sprintf("%s/repodata/repomd.xml", repo.URI)
 	_, err := os.Stat(repomdFile)
-	if err != nil {
-		err = helpers.Download(repomdURL, repomdFile)
+	if err != nil || update {
+		err := helpers.Download(repomdURL, repomdFile)
```
matthewrsj: Why shadow `err`?

matthewrsj: Also, if `update` you don't even need to do the stat right?

pkginfo/download_repo.go
```diff
@@ -85,7 +85,42 @@ func buildFilelistsURL(repo *Repo, workingDir string) (string, error) {
 	return fmt.Sprintf("%s/%s", repo.URI, path), nil
 }
 
-func buildPackageURLs(repo *Repo, flistsPath string) ([]string, error) {
+func compareVersionsAndRemoveStaleRPM(rpmURL, cacheDir string, cRPM *rpm.PackageFile) (int, error) {
+	newRPM, err := rpm.OpenPackageFile(rpmURL)
+	if err != nil {
+		return -2, err
```
matthewrsj: Why `-2`?

matthewrsj: looks like a negative error code, but this isn't C, there are better ways to do this in Go.

cmd/fetch.go
Gabib32: UInfo seems to only be in the fetch content. Do we want internal cmds calling the fetch logic or the download logic?

Gabib32: I'm all for putting it somewhere so that we don't have to pass it, I know we talked about potentially putting it in repo, but that it doesn't really relate to the repo info.

pkginfo/download_repo.go
Gabib32: Right, but I was trying to avoid calling helpers.Download more than once.

Gabib32: if !update {
		_, err := os.Stat(repomdFile)
	}
	if err != nil || update {
		err := helpers.Download(repomdURL, repomdFile)
		if err != nil {
			return "", err
		}
	}

matthewrsj: well err won't be defined in that second if block.

matthewrsj: uh, but yes something like that.

Gabib32: ohhh I see

Gabib32: I think because it needs to be redefined since its in the scope of that if statement?

pkginfo/download_repo.go
Gabib32: Okay, I did -2 because 0, 1, -1 are the only values so I wasn't sure what to use as a 'nil'

pkginfo/download_repo.go
```diff
@@ -34,12 +35,13 @@ var c *config.Config
...
 // construct the url using this value.
-func buildFilelistsURL(repo *Repo, workingDir string) (string, error) {
+func buildFilelistsURL(repo *Repo, workingDir string, update bool) (string, error) {
 	// download repomd.xml
 	repomdFile := filepath.Join(workingDir, "repomd.xml")
 	repomdURL := fmt.Sprintf("%s/repodata/repomd.xml", repo.URI)
+
 	_, err := os.Stat(repomdFile)
-	if err != nil {
+	if err != nil || update {
```
tmarcu: Are we actually trying to check if the repomdFile doesn't exist? If so should make this:
`	if os.IsNotExist(err) || update`

matthewrsj: I think it makes sense to re-download if stat fails for any reason. I'm not sure why else stat may fail though.

pkginfo/download_repo.go
```diff
@@ -85,7 +87,43 @@ func buildFilelistsURL(repo *Repo, workingDir string) (string, error) {
...
-func buildPackageURLs(repo *Repo, flistsPath string) ([]string, error) {
+// If a version changed from what was in the cache, return true, otherwise false
+func versionChanged(rpmURL, cacheDir string, cRPM *rpm.PackageFile) bool {
+	newRPM, err := rpm.OpenPackageFile(rpmURL)
+	if err != nil {
+		return true
+	}
+	// VersionCompare returns 1 if newRPM is more recent, -1 if cRPM is more recent
+	// and 0, if they're the same.
+	if rpm.VersionCompare(newRPM, cRPM) == 1 {
```
matthewrsj: this could be simplified to
```golang
return rpm.VersionCompare(newRPM, cRPM) == 1
```

pkginfo/import.go
```diff
@@ -19,18 +19,23 @@ import (
...
 
 // ImportAllRPMs imports all RPMs from a given repository. It populates the
 // passed repo with all RPMs imported.
-func ImportAllRPMs(repo *Repo) error {
+func ImportAllRPMs(repo *Repo, update bool) error {
 	var err error
 	var path string
-	if path, err = GetRepoFiles(repo); err != nil {
+
+	helpers.PrintBegin("Downloading upstream RPMs to cache")
```
matthewrsj: I'd rather these prints go outside the `ImportAllRPMs` function since this is pretty much meant to be used as a library function. I imagine something like this:
```golang
// in the caller
helpers.PrintBegin("Importing RPMs from %s", url)
err := pkginfo.ImportAllRPMs(&repo, false)
helpers.FailIfErr(err)
helpers.PrintComplete("RPMs imported ...")
```

pkginfo/import.go
```diff
@@ -19,18 +19,23 @@ import (
...
+	if path, err = GetRepoFiles(repo, update); err != nil {
 		return err
 	}
+	helpers.PrintComplete("RPM info downloaded to " + path)
 
+	helpers.PrintBegin("Importing RPMs to redis from cache")
```
matthewrsj: Let's not talk about `redis` since this is an internal implementation detail the user doesn't need to know about.

PR #58, opened 2018-07-19
helpers: convert Fail to FailIfErr

matthewrsj:
In almost every case helpers.Fail was wrapped in an if err != nil block.
Move this if block into the function itself and rename the function
FailIfErr.


cmd/bundle-verify.go
```diff
@@ -60,7 +60,7 @@ to correctly document data, and are defaulted to 0, and "clear", respectively.`,
 
 func runVerifyBundle(cmd *cobra.Command, args []string) {
 	if bundleFlags.repoURL == "" {
-		helpers.Fail(errors.New("must supply a --repourl argument"))
+		helpers.FailIfErr(errors.New("must supply a --repourl argument"))
```
tmarcu: Kinda weird to call FailIfErr( ) while passing in an explicit error, but for output consistency & less functions this is ok.

matthewrsj: My thought process exactly.

PR #64, opened 2018-07-25
Add functionality to collect all bundle packages

Gabib32:
This function collects all of the packages dependent by all
of the bundles within the bundlesdir.


matthewrsj:
+1, but how hard would it be to add a functional test for this critical package addition? :smirk:

Issue #45, opened 2018-07-03
Update pipcheck to create chroot


Gabib32:
If the user does not have a full chroot they would like to test pipcheck on, create on from the repo/url provided and run pipcheck against that.

Gabib32:
Done. Thoughts on pulling chroot creation out of the check into its own package? Would there ever be a use case for the canary check to create its own chroot from RPMs? @tmarcu @matthewrsj

matthewrsj:
hmm, that's a good question. But if it is needed we can open a different issue and discuss there. The issue opened here is fixed.

PR #65, opened 2018-07-25
Build full chroot from repo to run pipcheck against

Gabib32:
Install a specified repo as a full chroot, then run pipcheck against it, for instances when a chroot does not already exist.

TODO: Add a unit test for the new bundle library 'GetAllPackagesFromAllBundles" functionality.


Gabib32:
Resolves #45

cmd/pydeps.go
```diff
@@ -19,45 +19,60 @@ import (
...
 
 	p := pipFlags.path
 	if p == "" {
-		c, err := config.ReadConfig("")
-		helpers.FailIfErr(err)
-		p = filepath.Join(c.Mixer.MixWorkSpace, "update/image", pipFlags.version, "full")
+		p = filepath.Join(conf.Mixer.MixWorkSpace, "update/image", pipFlags.version, "full")
+	}
+
+	helpers.PrintBegin("Checking dnf and pip")
```
matthewrsj: you don't have a `helpers.PrintComplete()` and this should be fast enough you don't need a print at all.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
 	}
 }
 
+// Check both rpm and pip are installed on the system, to be used by the Pipcheck
+// and that the local chroot exists, or a repoURL is provided.
+func checkSystemRequirements() {
+	for _, tool := range []string{"rpm", "pip"} {
```
matthewrsj: Need to add `createrepo_c` to this list.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+func createFullChroot(path string) {
+	repo := pkginfo.Repo{
+		URI:     pipFlags.repoURL,
+		Name:    pipFlags.repoName,
+		Version: pipFlags.version,
+		Type:    "B",
+	}
+
+	helpers.PrintBegin("Importing RPMs from %s", repo.URI)
+	helpers.FailIfErr(pkginfo.ImportAllRPMs(&repo, pipFlags.update))
```
matthewrsj: You aren't being very consistent about how you are using `FailIfErr`, since it usually makes more sense to just call it on an `err` can we do that everywhere?

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+
+func createFullChroot(path string) {
+	repo := pkginfo.Repo{
+		URI:     pipFlags.repoURL,
+		Name:    pipFlags.repoName,
+		Version: pipFlags.version,
+		Type:    "B",
+	}
+
+	helpers.PrintBegin("Importing RPMs from %s", repo.URI)
```
matthewrsj: You should print a corresponding `PrintComplete`

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+	helpers.FailIfErr(pkginfo.ImportAllRPMs(&repo, pipFlags.update))
+
+	helpers.PrintBegin("Creating local repo for dnf install")
```
matthewrsj: Only use these helpers for long-running tasks. We don't want to clutter our output.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+	helpers.FailIfErr(helpers.RunCommandSilent("createrepo_c", repo.CacheDir))
+
+	// write dnf config file with repo information
+	dnfConf := "/tmp/dnf.conf"
```
matthewrsj: use https://golang.org/pkg/io/ioutil/#TempFile instead of hardcoding a path in `/tmp`.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+	helpers.FailIfErr(err)
+
+	// Get latest bundles directory and if a version is specified checkout that tag
+	helpers.FailIfErr(diva.GetLatestBundles(conf, ""))
+	if pipFlags.version != "0" {
+		tagArgs := []string{"-C", conf.Paths.BundleDefsRepo, "checkout", "tags/" + pipFlags.version}
+		helpers.FailIfErr(helpers.RunCommandSilent("git", tagArgs...))
+	}
+
+	helpers.PrintBegin("Get all bundle packages for install")
```
matthewrsj: This doesn't take long enough to need a print.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+	packages, err := bundle.GetAllPackagesForAllBundles(conf.Paths.BundleDefsRepo)
+	helpers.FailIfErr(err)
+
+	helpers.PrintBegin("Installing RPMs into chroot at %s", path)
```
matthewrsj: ok, this *does* take long enough for a print, but you need the `PrintComplete`

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+
+	dnfArgs := []string{"-c", dnfConf, "--installroot=" + path, "install", "-y",
+		"--releasever=" + pipFlags.version}
+
+	// install filesystem first
+	helpers.FailIfErr(helpers.RunCommandSilent("dnf", append(dnfArgs, "filesystem")...))
+	helpers.FailIfErr(helpers.RunCommandSilent("dnf", append(dnfArgs, packages...)...))
+
+	// clean up the temporary dnf config file
+	_ = os.Remove(dnfConf)
```
matthewrsj: throw this in the `defer` instead.

cmd/pydeps.go
```diff
@@ -66,12 +81,73 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+}
+
 // CheckPyDeps runs 'pip check' in a chroot at path
 func CheckPyDeps(path string) *diva.Results {
 	name := "Python dependencies"
 	desc := "run pip check in full build root to check for missing python requirements"
 	r := diva.NewSuite(name, desc)
 	r.Header(0)
+
+	helpers.PrintBegin("Running pipcheck")
```
matthewrsj: Please remove this, the user knows what check they called. Also pip check is fast.

cmd/pydeps.go
Gabib32: My only concern with TempFile, is that it will add a random sequence at the end of the filename and I"m not sure if DNF will correctly handle a config file that is dnf.conf21343423. I could create a tmpdir and put the dnf.conf inside of that instead?

matthewrsj: >  I could create a tmpdir and put the dnf.conf inside of that instead?

That's a good idea

bundle/bundle.go
```diff
@@ -273,3 +273,26 @@ func GetAllPackagesForBundle(name, bundlesDir string) ([]string, error) {
 	sort.Strings(sorted)
 	return sorted, nil
 }
+
+// GetAllPackagesForAllBundles gets every package used by the bundle definitions.
+// It returns an array of package names excluding any duplicates.
```
matthewrsj: Can you please either add a comment here that this returns a non-sorted slice or sort the slice before returning it (and also put a comment here that the returned slice is sorted)

cmd/pydeps.go
```diff
@@ -66,12 +81,90 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
 	}
 }
 
+// Check both rpm and pip are installed on the system, to be used by the Pipcheck
+// and that the local chroot exists, or a repoURL is provided.
+func checkSystemRequirements() {
+	for _, tool := range []string{"createrepo_c", "dnf", "pip"} {
+		err := helpers.RunCommandSilent(tool, "--version")
+		helpers.FailIfErr(err)
```
matthewrsj: To help with the defer issue, only call FailIfErr from the very top-level function. I think in this case `runCheckPyDeps` and everything else should return an error.

cmd/pydeps.go
```diff
@@ -66,12 +81,90 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+		URI:     pipFlags.repoURL,
+		Name:    pipFlags.repoName,
+		Version: pipFlags.version,
+		Type:    "B",
+	}
+
+	// download and import RPMs to cache and database
+	helpers.PrintBegin("Importing RPMs from %s", repo.URI)
+	err = pkginfo.ImportAllRPMs(&repo, pipFlags.update)
+	helpers.FailIfErr(err)
```
matthewrsj: See my comment about returning an error here instead.

cmd/pydeps.go
```diff
@@ -66,12 +81,90 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+
+	// clean the cache and metadata first
+	err = helpers.RunCommandSilent("dnf", "-c", dnfConf, "clean", "all")
+	helpers.FailIfErr(err)
+
+	dnfArgs := []string{"-c", dnfConf, "--installroot=" + path, "install", "-y",
+		"--releasever=" + pipFlags.version}
+
+	// install all bundle packages into full chroot, starting with filesystem
+	helpers.PrintBegin("Installing RPMs into chroot at %s", path)
```
matthewrsj: How about "preparing build root" instead of mentioning implementation details like "installing RPMs" and "chroot"

cmd/pydeps.go
```diff
@@ -66,12 +81,90 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+	err = helpers.RunCommandSilent("dnf", append(dnfArgs, "filesystem")...)
+	helpers.FailIfErr(err)
+
+	err = helpers.RunCommandSilent("dnf", append(dnfArgs, packages...)...)
+	helpers.FailIfErr(err)
+
+	helpers.PrintComplete("full Chroot creation completed")
```
matthewrsj: "done preparing build root" instead? You also capitalized Chroot for some reason.

cmd/pydeps.go
```diff
@@ -66,12 +81,90 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+}
+
 // CheckPyDeps runs 'pip check' in a chroot at path
 func CheckPyDeps(path string) *diva.Results {
 	name := "Python dependencies"
 	desc := "run pip check in full build root to check for missing python requirements"
 	r := diva.NewSuite(name, desc)
 	r.Header(0)
```
matthewrsj: You didn't change this but we might as well do `r.Header(1)` since we know we are only running one check.

bundle/bundle.go
```diff
@@ -273,3 +273,27 @@ func GetAllPackagesForBundle(name, bundlesDir string) ([]string, error) {
...
+		return nil, err
+	}
+
+	for _, bundle := range bundles {
+		for p := range bundle.AllPackages {
+			allPackages[p] = true
+		}
+	}
+
+	allPackageNames := []string{}
```
tmarcu: This might be over-optimizing, but all packages in all bundles is a lot and ever increasing, would be more efficient to do
`allPackageNames := make([]string, len(allPackages))`
Which will allocate an array of that size from the beginning so it doesn't have to resize as we then append all the values to it. 
ymmv :)

Gabib32: Ahh very smart and cool - thank you !

cmd/pydeps.go
```diff
@@ -66,12 +75,116 @@ func runCheckPyDeps(cmd *cobra.Command, args []string) {
...
+
+	repo := pkginfo.Repo{
+		URI:     "",
+		Name:    pipFlags.repoName,
+		Version: pipFlags.version,
+		Type:    "B",
+	}
+
+	// populate the repo information from the database
+	helpers.PrintBegin("Populating repo")
```
matthewrsj: Don't do here, but I realize now that we need to print this consistently across the codebase...

PR #68, opened 2018-07-27
Print more useful size change information

tmarcu:


cmd/check.go
```diff
@@ -110,8 +112,23 @@ func runBloatCheck(r *diva.Results, u diva.UInfo, args []string) error {
...
+		// Get an accurate percent change between versions.
+		// We have to break apart the float64 into integer & decimal parts so we
+		// can see if it increased or decreased.
+		intval, decimal := math.Modf(float64(toBundleSizes[bundle]) / float64(fromBundleSizes[bundle]))
+		if intval >= 1 {
+			decimal = (intval - 1 + decimal) * 100
+		} else {
+			decimal = (1 - decimal) * -100
+		}
+		pChange := fmt.Sprintf(" -> %3.2f%%", decimal)
```
matthewrsj: Since ` -> ` is always part of the printed test description can you move it to the string literal in the `fmt.Sprintf` below on L 131?

tmarcu:
Updated :) changed that print too many times so it got left over from a previous implementation.

cmd/check.go
```diff
@@ -110,8 +112,23 @@ func runBloatCheck(r *diva.Results, u diva.UInfo, args []string) error {
...
 		}
 		sizeDiff = toBundleSizes[bundle] - size
-		desc = bundle + " size did not change by more than " + strconv.FormatFloat(bloatFlags.warningCap, 'f', -1, 64) + "%"
-		ret := checkSize(&bundle, float64(sizeDiff), float64(size))
+		changeCap := bloatFlags.warningCap
+		_, ret := checkSize(&bundle, float64(sizeDiff), float64(size))
+
+		// Get an accurate percent change between versions.
+		// We have to break apart the float64 into integer & decimal parts so we
+		// can see if it increased or decreased.
```
icarus-sparry: I am puzzled by this comment and this code. Say the code goes from 10 to 12 bytes. Then you get a number (1.2) for the size change, and you subtract 1 and multiply by 100 to get a 20% change.

If the code goes from 10 bytes to 8 bytes, you get a number (0.8), you subtract 1 and multiply by 100 to get -20%.

If you just check the absolute value of the percentage change against changeCap, or else check that the percentage change is between changeCap and -changeCap will that not do, no need to get involved with math.Modf.

matthewrsj: @tmarcu

tmarcu: .. I'm not following your last paragraph. Yes, 10->12 is 20% increase, and 10->8 is -20%, which is what it should print. If it goes from 10 -> 8 you just re-ordered the math to .8 - 1 * 100 instead of 1 - .8 * -100, which make no difference mathematically. Maybe you can clarify what you want this to look like while still printing the actual % changed from ver1->ver2? :slightly_smiling_face:

tmarcu: @icarus-sparry pls recheck math now, simplified.

icarus-sparry: OK, the code used to look something like this (working from memory, the code is merged)

```
change=float(new_size)/float(old_size)
whole, fraction=math.Modf(change)
if (whole > 1) {
    percent=(whole- 1 fraction)*100
} else {
    percent = (1-whole - fraction) * -100
}
if (percent > 20) {
   fmt.Printf("Changed by too much)
}
```
My point is that it could just be

```
percent=(float(new_size)/float(old_size)-1.0)*100.0

if (percent > 20.0 || percent < -20.0) {
   fmt.Printf("Changed by too much)
}
```

PR #71, opened 2018-07-27
Add CacheDir in PopulateRepo if it doesn't exist

Gabib32:
If the repo.CacheDir is not set, populate a default. This
path needs to be used by some of the checks if they need
to use the actual RPMs.


pkginfo/repo.go
```diff
@@ -28,6 +31,20 @@ func PopulateRepo(repo *Repo) error {
 	defer func() {
 		_ = c.Close()
 	}()
+	conf, err := config.ReadConfig("")
```
matthewrsj: Should this take a config as an argument instead of always reading one, ignoring what was passed to the diva commandline?

Gabib32: ahh good call

PR #74, opened 2018-08-01
Add functionality to checkout a specific repo tag

Gabib32:
This runs a git checkout tags/<version> for the repo, so that
checks can be ran on more versions than latest. This is
currently being used to checkout different bundle definition
versions, but may be used by other repositories in the future.


diva/fetch.go
```diff
@@ -104,6 +104,32 @@ func FetchRepo(u UInfo) error {
 	return nil
 }
 
+// GetSpecificBundleTag clones the repo if it doesn't exist, and then checks
+// out the specific tag.
+func GetSpecificBundleTag(conf *config.Config, url, tag string) error {
```
matthewrsj: You can drop the `Specific` wording here since it is still implied in `GetBundleTag`

matthewrsj: Maybe `GetBundlesAtTag`?

diva/fetch.go
```diff
@@ -104,6 +104,32 @@ func FetchRepo(u UInfo) error {
...
+		helpers.PrintBegin("cloning latest bundle definitions")
+		err := helpers.CloneRepo(url, filepath.Dir(conf.Paths.BundleDefsRepo))
+		if err != nil {
+			return err
+		}
+		helpers.PrintComplete("bundle repo cloned to %s", conf.Paths.BundleDefsRepo)
+		return nil
+	}
+
+	helpers.PrintBegin("checking out bundle definitions for tag/%s", tag)
```
matthewrsj: This is very quick if the repo is already cloned. Please remove the prints.

matthewrsj:
@Gabib32 does this PR need to be updated now that #65 has been merged?

Gabib32:
Updated the PR to add the tag checkout to the pydeps command.

diva/fetch.go
```diff
@@ -97,6 +97,30 @@ func FetchRepo(u UInfo) error {
...
+		helpers.PrintBegin("cloning latest bundle definitions")
+		err := helpers.CloneRepo(url, filepath.Dir(conf.Paths.BundleDefsRepo))
+		if err != nil {
+			return err
+		}
+		helpers.PrintComplete("bundle repo cloned to %s", conf.Paths.BundleDefsRepo)
+		return nil
+	}
+
+	err := helpers.CheckoutRepoTag(conf.Paths.BundleDefsRepo, tag)
```
matthewrsj: just return here!

Gabib32: YAAASSS Good call :+1000000:

PR #75, opened 2018-08-01
Small commits to fix Fetch logic

Gabib32:
Have the default GetLatestVersion function return a string
value, and then provide a wrapper function that calls the
GetLatestVersion function but returns the value as a uint.


diva/fetch.go
```diff
@@ -53,26 +51,17 @@ func GetUpstreamInfo(conf *config.Config, upstreamURL string, version string, re
...
-	if err != nil {
-		return u, err
-	}
-	defer func() {
-		_ = resp.Body.Close()
-	}()
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return u, err
+	} else {
```
icarus-sparry: You could code this as
```
u.Ver = version
if u.Ver == "" {
      u.Ver, err = helpers.GetLatestVersion(u.URL)
```

to avoid the else.

Gabib32: :+1:

internal/helpers/helpers.go
```diff
@@ -293,19 +293,28 @@ func FailIfErr(err error) {
 
 // GetLatestVersion returns the version value at upstreamURL/latest or an error
 // if unable to do so.
-func GetLatestVersion(upstreamURL string) (uint, error) {
+func GetLatestVersion(upstreamURL string) (string, error) {
```
icarus-sparry: The comment **MUST** be fixed, to stop lint complaining. Once this is done Travis should pass and I can +1 it.

pkginfo/download_repo.go
```diff
@@ -247,7 +247,7 @@ func downloadAllRPMs(packages []string, workingDir string) error {
 // baseURL by first parsing the repo metadata. These packages are downloaded to
 // the c.CacheLocation/rpms/<version>/packages/ if they do not already exist
 // there.
-func GetRepoFiles(repo *Repo, update bool) (string, error) {
+func DownloadRepoFiles(repo *Repo, update bool) (string, error) {
```
icarus-sparry: Oops, last comment saying comment **MUST** be fixed was at wrong place 😦 This needs "GetRepoFiles" in the comment changed to "DownloadRepoFiles".

PR #76, opened 2018-08-02
Allow repo data to be updated when fetching

Gabib32:
**If you know of a better way to accommodate for the mash url, I would love ideas. This is the best way I could think of doing it without being like "If version == "mash"... which I know we don't want to hardcode mash in our code. Thoughts?**

When fetching repo/RPM data, allow an update option to be passed
that will prevent already downloaded RPMs from redownloading to
the cache. Additionally, if the conf.UpstreamURL is not used
as the repo.URI, then do not attempt to reformat the URL with
the version, but instead use the whole URL passed.


diva/fetch.go
```diff
@@ -75,14 +75,20 @@ func GetUpstreamInfo(conf *config.Config, upstreamURL string, version string, re
 
 // FetchRepo fetches the RPM repo at the u.URL baseurl to the local cache
 // location
-func FetchRepo(u UInfo) error {
+func FetchRepo(conf *config.Config, u UInfo) error {
```
matthewrsj: Don't pass the entire conf object to this function when all you need is the upstream url, just pass that.

Gabib32:
After talking with @bryteise I'm going to redo the logic on this and try again :)

matthewrsj:
mind filling in the peanut gallery?

PR #77, opened 2018-08-02
Change flags and bundle tag checkout

Gabib32:
Pluralize "bundle" and only pass in the config and tag string.
Use the config to get the bundleUrl information.


cmd/check.go
```diff
@@ -67,7 +67,7 @@ func runBloatCheck(r *diva.Results, u diva.UInfo, args []string) error {
 		u.Ver = args[0]
 	}
 
-	err = diva.GetBundleAtTag(conf, allFlags.bundleURL, u.Ver)
+	err = diva.GetBundlesAtTag(conf, u.Ver)
```
matthewrsj: need to set `conf.BundleDefsURL` to `allFlags.bundleURL` before calling this function.

cmd/pydeps.go
```diff
@@ -138,7 +138,7 @@ func createFullChroot(path string) error {
 	if pipFlags.version == "0" {
 		err = diva.GetLatestBundles(conf, "")
 	} else {
-		err = diva.GetBundleAtTag(conf, "", pipFlags.version)
+		err = diva.GetBundlesAtTag(conf, pipFlags.version)
```
matthewrsj: need to set `conf.BundleDefsURL` to `allFlags.bundleURL` before calling this function (I think, if it is available).

cmd/check.go
Gabib32: Do we really want to modify the config struct within the code?

matthewrsj: yeah, commandline arguments are just an override of the config so that's the closest way to handle it.

Gabib32: okay dokay

cmd/pydeps.go
Gabib32: I don't pass bundleDefsURL option to this command, but I"ll change that.

PR #78, opened 2018-08-03
Modify upstream info to update correctly

Gabib32:
The upstream info struct should only contain information
appropriate to the command being called. Therefore, create
an instance of the struct within the command call, and have
specific fields updated if they need to be, or are empty.


cmd/fetch.go
```diff
@@ -96,28 +97,35 @@ func init() {
...
 	fetchAllCmd.Flags().StringVarP(&fetchFlags.bundleURL, "bundleurl", "b", "", "URL from which to pull bundle definitions")
 	fetchAllCmd.Flags().StringVarP(&fetchFlags.upstreamURL, "upstreamurl", "u", "", "URL from which to pull update metadata")
 	fetchAllCmd.Flags().BoolVar(&fetchFlags.update, "update", false, "update pre-existing Repo")
+	fetchAllCmd.Flags().BoolVarP(&fetchFlags.recursive, "recursive", "r", false, "recursively fetch all content referenced in update metadata")
 
 	fetchBundlesCmd.Flags().StringVarP(&fetchFlags.bundleURL, "bundleurl", "b", "", "URL from which to pull bundle definitions")
 
 	fetchRepoCmd.Flags().StringVarP(&fetchFlags.version, "version", "v", "", "version from which to pull data")
+	fetchRepoCmd.Flags().StringVarP(&fetchFlags.upstreamURL, "upstreamurl", "u", "", "URL from which to pull update metadata")
+	fetchRepoCmd.Flags().BoolVar(&fetchFlags.update, "update", false, "update pre-existing Repo")
```
matthewrsj: `Repo` does not need to be capitalized here.

internal/helpers/helpers.go
```diff
@@ -296,6 +296,15 @@ func FailIfErr(err error) {
 	}
 }
 
+// GetVersionAsUint converts a version string into a uint
+func GetVersionAsUint(ver string) (uint, error) {
```
matthewrsj: I don't see the point of this function. It is just wrapping a single call to the std library `strconv.Atoi`, handling the error, and returning the cast value.

If you still want to keep the function, call it something more generic. There is nothing preventing this from being called on any numeric string but the name implies you are `Get`ting a `Version` when all you are doing is converting a string.

A better name for this function would be `ParseUint`...

cmd/fetch.go
Gabib32: :pick: e

Gabib32: :stuck_out_tongue_winking_eye:

internal/helpers/helpers.go
tmarcu: https://golang.org/pkg/strconv/#ParseUint

Gabib32: Yikes, yeah I have no idea what I was thinking with this.. /me fixing

Gabib32: :( ParseUint is uint64, so still needs the masking to uint

PR #79, opened 2018-08-06
WIP - do not merge Enable runtime tracing

icarus-sparry:
Think of this as a technology preview. Add a "--trace" just after "diva" and then run "go tool trace --http:10159 trace.out" to see where time is going - open web browser to http://localhost:10159/.

Add --trace flag globally, to support "go tool trace".

Signed-off-by: Icarus Sparry <icarus.w.sparry@intel.com>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


tmarcu:
N/A

PR #81, opened 2018-08-21
Add canary check to diva

tmarcu:
This check takes in two full chroot directories, and reads the dynamic
symbols section of every valid ELF file to confirm there is a
`__stack_chk_fail` symbol signifying a stack canary is present. If it is
missing from either the old or new file, it is an error, and if the old
file does not exist, the new one must still contain it.


Gabib32:
Put this in its own file - like canary.go or something, not with the check.go file!

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+		return false, err
+	}
+	if ident[0] != '\x7f' || ident[1] != 'E' || ident[2] != 'L' || ident[3] != 'F' {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func readCanary(r *diva.Results, u diva.UInfo, args []string) error {
```
matthewrsj: looks like `u diva.UInfo` is unused in this function.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+	oldfile := args[0]
+	newfile := args[1]
+
+	var err error
+	prev, err := os.OpenFile(oldfile, os.O_RDONLY, 0)
+	if os.IsNotExist(err) {
+		// We possibly have a new file, so mark as nil for now
+		prev = nil
```
matthewrsj: will `prev` not be assigned `nil` anyways by `os.OpenFile` `if os.IsNotExist(err)`?

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+	} else if err != nil {
+
```
matthewrsj: extra newline?

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+	defer func() {
+		_ = newFile.Close()
+	}()
+
+	var newElf *elf.File
+
+	// Don't fail on non-elf files, so check before trying to parse
+	var ret bool
+	ret, err = isElf(newFile)
+	if err != nil {
```
matthewrsj: collapse the next few lines into
```go
if !ret || err != nil {
    return err
}
```

matthewrsj: Should you move this step earlier and re-order a bit so you do as few reads as possible if the new file isn't an ELF file?

1. read new file
2. check if new file is ELF, return if not
3. read old file

that way you only read the old file if the new file is an ELF.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+		return err
+	}
+	// It's just not an elf file, so ignore trying to check for canary
+	if !ret {
+		return nil
+	}
+
+	// Get elf specific struct to read symbols from
+	newElf, err = elf.NewFile(newFile)
```
matthewrsj: This should be moved before reading the old file as well.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+	var returnErr string
+	var prevElf *elf.File
+	if prev != nil {
+		ret, err = isElf(prev)
+		if err != nil {
+			return err
+		}
+		// It's just not an elf file, so ignore trying to check for canary
+		if !ret {
+			return nil
```
matthewrsj: Don't return here, you know the new file is an ELF file and still needs to be checked.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+		}
+		prevElf, err = elf.NewFile(prev)
+		if err != nil {
+			return err
```
matthewrsj: Same, this shouldn't return so the new file is still checked.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+		}
+		if err = hasCanary(oldfile+"\n", prevElf); err != nil {
+			returnErr += err.Error()
```
matthewrsj: It isn't an error if the old file doesn't have a canary. It is an error if new file doesn't have a canary and old file does, it is a warning if both old and new files do not have canary.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+		}
+	}
+	if err = hasCanary(newfile, newElf); err != nil {
+		returnErr += err.Error()
+	}
+	if returnErr != "" {
+		return errors.New(returnErr)
```
matthewrsj: This will report an error if the old file was missing a canary but the new file added one in.

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+	}
+	return nil
+}
+
+func getFiles(path string, info os.FileInfo, err error) error {
+	// Only add valid regular files to the list to check
+	if info.Mode().IsRegular() {
+		filesList = append(filesList, path)
```
matthewrsj: using a global here is kinda gross when you could have getFiles return a callback instead. See the `Passing Arguments` header here: https://xojoc.pw/justcode/golang-file-tree-traversal.html

matthewrsj:
```go
func getFiles(files []string) os.WalkFunc {
    return func(path string, info os.FileInfo, err error) error {
        if info.Mode().IsRegular() {
            files = append(files, path)
        }
    }
}

func runCanaryCheck(args arguments) (errs []string) {
    // process args
    files := []string{}
    err := filepath.Walk(newBuild, getFiles(files))
    // operate on files
}
```

cmd/check.go
```diff
@@ -14,14 +14,191 @@
...
+func isElf(r io.ReaderAt) (bool, error) {
+	var ident [16]uint8
+	if _, err := r.ReadAt(ident[0:], 0); err != nil {
+		// If this fails with EOF it means the header doesn't even exist in the file
+		if err == io.EOF {
+			return false, nil
+		}
+		return false, err
+	}
+	if ident[0] != '\x7f' || ident[1] != 'E' || ident[2] != 'L' || ident[3] != 'F' {
```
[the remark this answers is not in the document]
tmarcu: This is directly pasted from the "debug/elf" which works well currently, so I didn't want to change more than the return values in it. Also you can't compare slices in Go, unless you use reflect.DeepEqual() or bytes.Equal() which I don't wanna pull in just for this.

cmd/check.go
tmarcu: woops, artifact from when I had the results API in this file instead of in the main check function. Updating!

cmd/check.go
tmarcu: hmm look like it should return nil according to the codes https://golang.org/src/os/file_unix.go

cmd/check.go
tmarcu: No it makes sense in the order it is I think:
1) If new file doesn't exist, it's no longer in the distro so we don't care
1.5) If new file isn't an ELF file anymore, it's essentially not in the distro so we don't care; i'm not sure when this would ever happen anyway.

And yes I agree it should be moved first so we can exit early before ever opening old file :man_dancing:

cmd/check.go
tmarcu: hmmm from our conversations last week the consensus was that any file not having a canary is a failure.

cmd/check.go
matthewrsj: > No it makes sense in the order it is I think:

Wait, the order you suggested is the order I suggested. So we agree?

cmd/canary.go
```diff
@@ -0,0 +1,186 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/clearlinux/diva/diva"
+)
+
+const canaryString = "__stack_chk_fail"
+
+var chirpCmd = &cobra.Command{
```
matthewrsj: lol

cmd/check.go
```diff
@@ -14,7 +14,9 @@
 
 package cmd
 
-import "github.com/spf13/cobra"
+import (
```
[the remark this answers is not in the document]
tmarcu: It used to have like 8 imports which are now in canary.go :)

cmd/canary.go
```diff
@@ -0,0 +1,204 @@
...
+			r.Diagnostic(fmt.Sprint(len(errs)) + " Canaries Missing:\n" + strings.Join(errs, "\n"))
+		}
+		r.Ok(len(warnings) == 0, "No issues since last build, all existing ELF issues resolved")
+		if len(warnings) > 0 {
+			r.Diagnostic(fmt.Sprint(len(warnings)) + " Warnings:\n" + strings.Join(warnings, "\n"))
+		}
+
+		if chirpFlag {
+			for i := 0; i < len(errs); i++ {
+				fmt.Printf("chirp ")
```
matthewrsj: :baby_chick:

tmarcu: :bird:

Gabib32: ❤️

cmd/canary.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+	return findCanary(newfile, newElf)
+}
+
+func readCanary(oldfile, newfile string) (string, error) {
+	var err error
+	// If newfile exists AND has a canary, don't waste time parsing previous
+	if err = getNewFileCanary(newfile); err == nil {
```
matthewrsj: Nice logic here

cmd/canary.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	defer func() {
+		_ = prev.Close()
+	}()
+
+	// Check both files if they exist to make sure the canary exists and was
+	// not lost between the two build versions.
+	var prevElf *elf.File
+	// Check if old file has canary if new doesn't
+	if prev != nil && returnErr != "" {
```
matthewrsj: At this point won't `returnErr` always be `!= ""` since you set it to a non-nil `err.Error()` above?

cmd/canary.go
```diff
@@ -0,0 +1,204 @@
...
+		isELFfile, err = isElf(prev)
```
matthewrsj: You could totally generalize `getNewFileCanary` to `getFileCanary` and just call it again here.

tmarcu:
Ok V4 up, thanks for the really good reviews btw.

cmd/canary.go
```diff
@@ -0,0 +1,168 @@
...
+	fileELF, err = elf.NewFile(fd)
+	// If we get an err here, it IS an elf file but couldn't be parsed.
+	// We need to check if the previous was an elf file and had a canary.
+	if err != nil {
+		return err
+	}
+	return findCanary(file, fileELF)
+}
+
+func readCanary(oldfile, newfile string) (string, error) {
```
matthewrsj: This func looks great now

cmd/canary.go
```diff
@@ -0,0 +1,168 @@
+package cmd
+
+import (
+	"debug/elf"
+	"fmt"
+	"github.com/spf13/cobra"
```
Gabib32: stupid formatting thing, but is this supposed to be in its own section?

matthewrsj: good call

cmd/check.go
```diff
@@ -24,4 +24,5 @@ var checkCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(checkCmd)
+	checkCmd.AddCommand(chirpCmd)
```
Gabib32: For consistency, all of the other checks have their own init functions within their files that have checkCmd.AddCommand(). We should keep it the same and either move this, or move the other AddCommand calls - whichever you guys think is better!

cmd/canary.go
```diff
@@ -0,0 +1,186 @@
...
+			errs = append(errs, err.Error())
+		}
+	}
+	return errs
+}
+
+var chirpFlag bool
+
+func init() {
+	chirpCmd.Flags().BoolVar(&chirpFlag, "chirp", false, "Print number of errors in chirps")
```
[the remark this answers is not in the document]
Gabib32: hahahaha this is so great! Definitely using this with every call :dancer:

cmd/canary.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			errs = append(errs, errString)
+		} else if errString != "" {
+			warnings = append(warnings, errString)
+		}
+	}
+	return errs, warnings
+}
+
+var chirpFlag bool
```
Gabib32: Maybe this and the init func can be moved to the top for consistency and going with the cmd calls - but I may just be being too nitpicky, so feel free to ignore ;)

cmd/canary.go
```diff
@@ -0,0 +1,168 @@
...
+	var isELFfile bool
+	isELFfile, err = isElf(fd)
+	if err != nil || !isELFfile {
+		return err
+	}
+
+	var fileELF *elf.File
+	fileELF, err = elf.NewFile(fd)
+	// If we get an err here, it IS an elf file but couldn't be parsed.
+	// We need to check if the previous was an elf file and had a canary.
```
Gabib32: +1

cmd/canary.go
matthewrsj: @tmarcu are you going to update this or are you going to ignore it?

cmd/canary.go
matthewrsj: Oh also this @tmarcu

tmarcu:
@matthewrsj updated formatting comments

matthewrsj:
oman I was gonna merge this but travis

matthewrsj:
Please rebase against master so the travis failures go away

PR #83, opened 2018-08-30
Large refactor

Gabib32:
This is a large refactor that will probably require many reviews. I still need to modify some of the arguments to pass copies instead of pointers everywhere, and fix the bundle library with its unit tests. Other than that, this is kind of the way I was going to format things easier for a user. The difficulty was that the fetch and check arguments needed to be relatively similar, so that the db keys and whatnot were unique - especially for when users want to use "latest" or "mash" for the versions.


internal/helpers/helpers.go
```diff
@@ -323,3 +324,13 @@ func GetLatestVersionUint(upstreamURL string) (uint, error) {
 	ver, err := strconv.ParseUint(versionString, 10, 32)
 	return uint(ver), err
 }
+
+// HashmapToSortedSlice converts a hashmap to a sorted slice
+func HashmapToSortedSlice(hashmap map[string]bool) ([]string, error) {
```
matthewrsj: \o/ +1

bundle/bundle_test.go
```diff
@@ -488,76 +488,76 @@ func TestAllBundlesSet(t *testing.T) {
...
-	for _, tc := range testCases {
-		t.Run(tc.bundleName, func(t *testing.T) {
-			actualIncludes, err := GetIncludesForBundle(tc.bundleName, testData.testdir)
-			if err != nil || !reflect.DeepEqual(tc.expectedIncludes, actualIncludes) {
-				t.Error(deep.Equal(tc.expectedIncludes, actualIncludes))
-			}
-		})
-	}
-}
+// func TestGetIncludesForBundle(t *testing.T) {
```
matthewrsj: Why is everything commented out?

Gabib32: lol because I changed the function being tested, and I wasn't sure whether to keep the old logic too, or replace completely. I'm leaning towards the latter. (AKA the test would need to change)

PR #84, opened 2018-09-04
Fix travis issues

Gabib32:
Fix travis issues with lint


internal/helpers/helpers.go
```diff
@@ -203,16 +203,16 @@ func RunCommandOutput(cmdname string, args ...string) (*bytes.Buffer, error) {
 	err := cmd.Run()
 	if err != nil {
 		var buf bytes.Buffer
-		fmt.Fprintf(&buf, ": failed to execute %s", strings.Join(cmd.Args, " "))
+		_, _ = fmt.Fprintf(&buf, ": failed to execute %s", strings.Join(cmd.Args, " "))
```
tmarcu: If we're writing to a custom buffer here, do we want to actually check that the right number of bytes was written instead of throwing away the returns?

Gabib32: And then fail on a failure?

Gabib32: Works for me if we care enough to error handle the error handler!

PR #85, opened 2018-09-04
Restructure bundle behavior

Gabib32:
There has to be a better way of doing this.... I feel like this library is a mess. Suggestions?


Gabib32:
I have not modified any of the other diva code - first finding a solid solution for the bundle library before modifying the rest of the code to use it.

PR #89, opened 2018-09-07
Large refactor for data storage

Gabib32:
This PR contains multiple changes, including:

- Splitting database store and load functionalities into their own files
- Store bundle Information and Repo information to the database, which also splits the command calls to appropriately be fetch and check. The check calls no longer download any bundle or repo data, and the fetch calls download, update the cache, and import to the database. 
- Add methods to the bundle object for package and include modifications
- Update pundle delete detection and add bundle delete detection, so that the tag comparing to HEAD is the version of the repo checked out, and not always latest (upstream does not have a latest tag).
- Speed up the cache updating by using a map for searching and comparing RPM names, as opposed to downloading the RPM to the go-rpm library and comparing that way.


internal/helpers/helpers.go
```diff
@@ -323,3 +324,13 @@ func GetLatestVersionUint(upstreamURL string) (uint, error) {
...
 }
+
+// HashmapToSortedSlice converts a hashmap to a sorted slice
+func HashmapToSortedSlice(hashmap map[string]bool) ([]string, error) {
+	sorted := make([]string, 0, len(hashmap))
+	for p := range hashmap {
+		sorted = append(sorted, p)
+	}
+	sort.Strings(sorted)
+	return sorted, nil
```
matthewrsj: any reason not to `return sort.Strings(sorted), nil` here? Not a big deal.

Gabib32: lol I think you've asked this before ;) I think sort.Strings(sorted) doesn't return what we would want it to. It errors out I believe

matthewrsj: Oh that's right, it doesn't actually return the slice, sorry. +1

cmd/bundle-verify.go
```diff
@@ -90,29 +101,16 @@ func runVerifyBundle(cmd *cobra.Command, args []string) {
...
-		var singleBundle *bundle.Definition
-		singleBundle, err = bundle.GetDefinition(bundleFlags.bundle, conf.Paths.BundleDefsRepo)
-		if singleBundle != nil {
-			bundles[singleBundle.Name] = singleBundle
-		}
-	}
-
+func checkBundleDefinitions(result *diva.Results, bundleInfo *pkginfo.BundleInfo, bundleName string) error {
+	err := pkginfo.PopulateBundles(bundleInfo, bundleName)
 	result.Ok(err == nil, "no include loops")
```
matthewrsj: Looking at your code this doesn't actually validate that there are no include loops since that check happens during fetch. We discussed offline better approaches.

cmd/bundle-verify.go
```diff
@@ -51,35 +58,39 @@ var verifyBundlesCmd = &cobra.Command{
...
+		Latest:      bundleFlags.latest,
+		RPMType:     bundleFlags.rpmType,
+		BundleCache: bundleFlags.bundleCache,
+		RPMCache:    bundleFlags.rpmCache,
 	}
+	pkginfo.UpdateConfigInstance(conf, u)
 
-	helpers.PrintBegin("Populating repo")
-	err := pkginfo.PopulateRepo(&repo, conf.Paths.CacheLocation)
+	bundleInfo, err := pkginfo.NewBundleInfo(conf, &u)
```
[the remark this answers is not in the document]
Gabib32: I agree - I am down with figuring out something a lot better

bundle/bundle_test.go
```diff
@@ -488,76 +488,76 @@ func TestAllBundlesSet(t *testing.T) {
...
-	for _, tc := range testCases {
-		t.Run(tc.bundleName, func(t *testing.T) {
-			actualIncludes, err := GetIncludesForBundle(tc.bundleName, testData.testdir)
-			if err != nil || !reflect.DeepEqual(tc.expectedIncludes, actualIncludes) {
-				t.Error(deep.Equal(tc.expectedIncludes, actualIncludes))
-			}
-		})
-	}
-}
+// func TestGetIncludesForBundle(t *testing.T) {
```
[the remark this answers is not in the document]
Gabib32: Fixed it to be a working test. Thank you for the reminder :)

bundle/bundle.go
```diff
@@ -282,3 +283,16 @@ func GetAllPackagesForAllBundles(bundlesDir string) ([]string, error) {
...
+// and returns a sorted slice of all unique packages.
+func GetAllPackagesFromAllBundleDefinitions(bundles []*Definition) ([]string, error) {
+	allPackages := make(map[string]bool)
+
+	for _, bundle := range bundles {
+		for p := range bundle.AllPackages {
+			allPackages[p] = true
+		}
+	}
+	return helpers.HashmapToSortedSlice(allPackages)
```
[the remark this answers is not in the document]
Gabib32: what about duplicates? Also I changed a lot of the bundle logic in the following commit, so some of this stuff is gone, should I merge that commit into this one?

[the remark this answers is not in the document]
matthewrsj: ah, forgot there may be duplicates across maps. No this is good. No please don't merge commits. I'm not quite done reviewing this one yet.

pkginfo/download_repo.go
```diff
@@ -86,40 +86,27 @@ func buildFilelistsURL(repo *Repo, workingDir string, update bool) (string, erro
...
-				cRPM.Version(),
-				cRPM.Release(),
-				cRPM.Architecture(),
-			)
-			if err := os.Remove(staleRPM); err != nil {
-				return true
-			}
-		}
+	cachedRPMURL := filepath.Join(rpmCache, cachedRPM)
+	if err := os.Remove(cachedRPMURL); err != nil {
```
matthewrsj: you have 
```go
if err := os.Remove(cachedRPMURL); err != nil {
    return true
}
return true
```
Do you not care about the error? You are returning true no matter what so you can do this:
```go
_ = os.Remove(cachedRPMURL)
return true
```

cmd/bundle-verify.go
```diff
@@ -164,3 +167,30 @@ func checkIfPundleDeletesExist(result *diva.Results) error {
...
+		"log", "--diff-filter=D", tag+"..HEAD", "--summary")
+	if err != nil {
+		return err
+	}
+
+	scanner := bufio.NewScanner(output)
+	for scanner.Scan() {
+		// typically lines are formatted like: ' delete mode 100644 bundles/<bundleName'
+		// ^\sdelete\smode\s[0-9]*\sbundles/.*$
+		matches := regexp.MustCompile(`^\sdelete.*bundles/.*$`).FindStringSubmatch(scanner.Text())
```
matthewrsj: You are recompiling this regexp every iteration of this loop, yet the regexp doesn't actually change. Please compile the regexp before the loop and only call the `rxp.FindStringSubmatch(scanner.Text())` within the loop.

matthewrsj: Shoot I noticed the same issue in pundle deletes. Please update that one too.

pkginfo/types.go
```diff
@@ -183,12 +184,24 @@ type BundleInfo struct {
...
 func updateBundleInfo(bundleInfo *BundleInfo, u *UInfo, conf *config.Config) error {
-	return bundleInfo.Repo.Update(&bundleInfo.Repo, u, conf)
+	err := bundleInfo.Repo.Update(&bundleInfo.Repo, u, conf)
+	if err != nil {
+		return err
+	}
+
+	bundleInfo.Tag = bundleInfo.Version
+	// if bundle version is not a valid version number, or is 0, use latest
+	_, isNum := strconv.Atoi(bundleInfo.Version)
```
Gabib32: Rename to err

pkginfo/types.go
```diff
@@ -58,3 +173,84 @@ type File struct {
...
+}
+
+// ManifestInfo contains all fields from the BundleInfo struct, along with the
+// Version as an Uint instead of a string, a minVer, and an update function for
+// all embedded datatypes.
+type ManifestInfo struct {
+	BundleInfo
+	SecondBundle BundleInfo
+	VerAsUint    uint
+	MinVer       uint
```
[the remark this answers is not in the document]
Gabib32: The recursive arg, so when fetching data, takes from whatever the minversion is (if recursive is passed, that's 0 - otherwise its the current version). Also for the update content checks. Everything passes in a minVer

pkginfo/load_redis_test.go
```diff
@@ -86,3 +88,62 @@ func TestGetRepoRedis(t *testing.T) {
...
+	}
+
+	for _, bun := range bundleInfo.BundleDefinitions {
+		if bun.Name != bundleName {
+			t.Errorf("expected bundle name %s, but got %s", bundleName, bun.Name)
+		}
+		if bun.Header.Capabilities != "testcap" {
+			t.Errorf("expected bundle Header Capabilities testcap, but got %s", bun.Header.Capabilities)
+		}
+		if _, ok := bun.AllPackages["slice_yeah"]; ok != true {
```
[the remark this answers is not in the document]
Gabib32: HAHAHA This will NOT be the last time I do something like this :whale2:

pkginfo/download_repo.go
Gabib32: ooo good call (I don't care - which I thiiiink is okay)

cmd/bundle-verify.go
Gabib32: Oh GOOD call! Thank you :bowing_man:

PR #91, opened 2018-09-11
Download and import SRPMs and Binary RPMs by default

Gabib32:
By default download both the binaries and SRPMs to the cache
and import both into the database. If a user wants to specify
only one RPM type, they can pass --source or --binary depending
on whichever they desire.


Gabib32:
Good call on the comments!!

The filepath.Join strips one of the slashes from the url prefix (https:// -> https:/) resulting in an error to download from the url. Should I use net/url package?

tmarcu:
Ahh I missed that, I wouldn’t pull in another package just for that, I think it looks/works fine with regular sprintf. So ignore my earlier comment where it doesn’t make sense to change it :) 
You could write
`Sprintf(“%s/%s”, repo.UpstreamURL, filepath.Join(...))` ...but it looks uglier than what you already have 🙂

PR #95, opened 2018-09-24
Add subcommands under main check command

tmarcu:

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -24,5 +24,9 @@ var checkCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(checkCmd)
+	checkCmd.AddCommand(bloatCheckCmd)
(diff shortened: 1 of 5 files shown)


cmd/check.go
```diff
@@ -24,5 +24,9 @@ var checkCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(checkCmd)
+	checkCmd.AddCommand(bloatCheckCmd)
 	checkCmd.AddCommand(chirpCmd)
+	checkCmd.AddCommand(pyDepsCmd)
+	checkCmd.AddCommand(ucCmd)
+	checkCmd.AddCommand(verifyBundlesCmd)
```
matthewrsj: There are other commands that should be added here (like fetch), but we can do that as a different PR.

PR #99, opened 2018-10-03
Organize fetch and checks and import bundles

Gabib32:
The architecture for DIVA is to separate the fetching
of data from the different checks. This has been
accomplished in this commit.

The repo and bundle data, when fetched, is downloaded and
imported into the databse. The update content still needs
this functionality. There are different data structs that
populate the needed information from the database, so the
check commands are not downloading any content.


bundle/bundle.go
```diff
@@ -50,8 +48,69 @@ type Definition struct {
...
 }
 
-// Set is a map of bundle names to their definition
-type Set map[string]*Definition
+// DefinitionsSet is a map of bundle names to their definition
+type DefinitionsSet map[string]*Definition
+
+//
+// // Definitions is a slice of Definition objects
+// type Definitions []*Definition
```
matthewrsj: remove commented-out code please

bundle/bundle.go
```diff
@@ -50,8 +48,69 @@ type Definition struct {
...
+
+	allIncludes := make(map[string]bool)
+	for _, bundle := range defs {
+		for j := range bundle.Includes {
+			allIncludes[j] = true
+		}
+	}
+	var err error
+	if len(allIncludes) == 0 {
+		err = fmt.Errorf("Unable to get all includes, set is empty")
```
matthewrsj: are we assuming that there will always be one include (like `os-core`)?

matthewrsj: But what about os-core? that doesn't have any includes and that isn't an error.

matthewrsj: good test case to add ;)

bundle/bundle.go
```diff
@@ -50,8 +48,69 @@ type Definition struct {
...
+
+	return allIncludes, err
+}
+
+// GetAllPackages returns all of the dependent packages recursively among all
+// bundle definitions. If a bundleName is passed, only the recursive packages
+// to that bundle will be returned.
+func (defs DefinitionsSet) GetAllPackages(bundleName string) (map[string]bool, error) {
+	if bundleName != "" {
+		return defs[bundleName].AllPackages, nil
```
matthewrsj: \o/

bundle/bundle.go
```diff
@@ -50,8 +48,69 @@ type Definition struct {
...
+	allPackages := make(map[string]bool)
+	for _, bundle := range defs {
+		for j := range bundle.AllPackages {
+			allPackages[j] = true
+		}
+	}
+
+	var err error
+	if len(allPackages) == 0 {
+		err = fmt.Errorf("Unable to get all packages, set is empty")
```
matthewrsj: is it invalid to have a bundle definition with no packages (and no includes)?

bundle/bundle.go
Gabib32: +1 good catch

Gabib32: ahhhh but os-core will include itself ;)

Gabib32: So boom, it does work :stuck_out_tongue_closed_eyes:

bundle/bundle.go
Gabib32: I don't even know why I bother with this, instead of in the code just calling the defs[name].includes ... merp :woman_shrugging:

Gabib32: ^ that's a shrug btw lol

bundle/bundle.go
Gabib32: Since they all include themselves, I would think yes. @tmarcu thoughts?

bundle/bundle.go
matthewrsj: hmm, I forgot this list is supposed to also list the parent bundle. That's confusing, but ok.

bundle/bundle.go
matthewrsj: Not sure I agree with this. I think if a user wants to define an empty OS (no packages even in os-core) they should be allowed to. At least they shouldn't be prohibited by the bundle library. bundle *validation* on the other hand should probably complain.

bundle/bundle.go
matthewrsj: ... that's a good point lol

bundle/bundle_test.go
```diff
@@ -588,49 +593,64 @@ func TestGetAllPackagesForOneBundle(t *testing.T) {
...
 	}
 
 	for _, tc := range testCases {
 		// add os-core packages to all bundles
 		for pkg := range core.AllPackages {
-			tc.expectedPackages = append(tc.expectedPackages, pkg)
+			tc.expectedPackages[pkg] = true
+		}
+
+		for _, pkg := range tc.expectedSlice {
```
matthewrsj: wait so is the only reason you have expectedSlice so you can add it to the expectedPackages map? Why not just define the expectedPackages as a literal above and add the core.AllPackages like you do?

Gabib32: I guess I can, I just already had the slices programmed, so it saved me some of that hardcoding time :) But I'll do that so its consistent throughout

Gabib32: lol that took like 2 minutes I was beign so lazy

Issue #80, opened 2018-08-07
Add GetAllSRPMs function to pkginfo


matthewrsj:
Useful for find2 conflicts and family.

Gabib32:
Added in #100. The download/fetch repo option automatically gets both RPMs and SRPMs, unless the boolean options "--source" or "--binary" are passed.

PR #100, opened 2018-10-03
Phasethree

Gabib32:


cmd/bloat.go
```diff
@@ -20,22 +20,82 @@ import (
...
+	Args: cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+
+		// set Ver to be the first arg passed, in case only one arg is passed
+		u := config.UInfo{
+			Ver:     args[0],
+			Latest:  bloatFlags.latest,
+			MixName: bloatFlags.mixName,
+		}
+		config.UpdateConfigInstance(conf, u)
```
tmarcu: I'm not sure you want to do this before you figure out which argument should come first, do you?

Gabib32: Actually, I don't think this should be called at all... hmmm I think the UpdateConfigInstance may only be needed for the fetching calls, since those would be the only things modifying urls/cache locations (since most everything will be imported from redis). Good call. :)

Gabib32: The entire Uinfo/flags/etc thing got me all spun around - if you have a better way of handling it I would LOVE to hear it, orrr if you think we should keep the old way? I'm open to anything

pkginfo/download_bundles.go
```diff
@@ -22,13 +22,40 @@ import (
...
+		return true
+	}
+
+	currentRepo, err := helpers.RunCommandOutput("git", "-C",
+		bundleInfo.BundleCache, "config", "--get", "remote.origin.url")
+	if err != nil {
+		return true
+	}
+
+	if strings.TrimSpace(currentRepo.String()) != bundleInfo.BundleURL {
```
matthewrsj: nice, didn't even consider this.

pkginfo/download_bundles.go
```diff
@@ -22,13 +22,40 @@ import (
...
+		err := os.RemoveAll(bundleInfo.BundleCache)
```
matthewrsj: scares me to remove git repositories, but I guess it's ok bc it is in the defined `bundleInfo.BundleCache` location.

pkginfo/download_bundles.go
Gabib32: Right! I didn't either until I actually ran into it. Yaaay

pkginfo/download_bundles.go
Gabib32: we could exit out with an error instead? Now allow them to download the bundles?

matthewrsj: Yeah that's probably better actually. Make them fix their stuff instead of us violently wiping a git repo.

matthewrsj: as long as the error is descriptive enough.

matthewrsj:
Fixes #80

PR #103, opened 2018-10-04
Do Not Merge - just the start to the find2 check

Gabib32:
In case you guys wanted to take a look at it at all, this is all I started for it - nothing special at all.


matthewrsj:
I don't really like the command `diva check rpms` since there are several individual checks that will be happening.
* `diva check conflicts`
* `diva check dependencies`

are the basic cases.

Gabib32:
So we should split the checks, so the user has to know all the different checks to run relating to the RPMs?

Gabib32:
I agree I don't like the name/command check rpms +1

matthewrsj:
> So we should split the checks, so the user has to know all the different checks to run relating to the RPMs?

Yeah sounds reasonable to me. They shouldn't necessary have to know that these specific checks are even looking at RPMs.

Gabib32:
@tmarcu This was a start I made for the dependency check - I will 0% be offended if you throw it out and start fresh ( I would recommend that actually). But if you wanted a place to start, or to see what I was attempting you're welcome to take a glance :)

PR #107, opened 2018-10-30
Update bloat check to use database

Gabib32:
Now that the manifest data and files are stored in the
database, modify the check to use that data instead of
populating it from the cache location.


matthewrsj:
I'll clarify that +1: the "Update bloat check to use database" is what lgtm

Gabib32:
@tmarcu do you want to look this one over before I merge it at all?

PR #111, opened 2018-11-07
Getting repo URI should not return error

Gabib32:
Getting the repo URI is only necessary for downloading and
fetching data, so during check commands it does not need to
be set. Instead of erroring if it cannot be set, use the
empty string, and print the errors.


pkginfo/types.go
```diff
@@ -98,15 +98,9 @@ func (repo *Repo) updateRepo(u *config.UInfo) error {
 	}
 
 	if repo.Type == "SRPM" {
-		err = getRepoURI(repo, "source/SRPMS")
-		if err != nil {
-			return err
-		}
+		getRepoURI(repo, "source/SRPMS")
```
tmarcu: If this is the only places we look I think this is OK, will we ever care about the return or not?

pkginfo/types.go
```diff
@@ -76,11 +76,11 @@ func getRepoURI(repo *Repo, loc string) error {
 		_, err := helpers.CheckStatus(url)
 		if err == nil {
 			repo.URI = url
-			return nil
 		}
 		errs = append(errs, err)
 	}
-	return fmt.Errorf("Unable to get valid rpm URI: %v", errs)
+	repo.URI = ""
+	fmt.Println(fmt.Errorf("Unable to get valid rpm URI. Errors: %v", errs))
```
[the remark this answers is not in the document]
matthewrsj: why are we printing an error from an internal function anyways?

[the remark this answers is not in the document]
matthewrsj: If it isn't really an error but you still want it to be printed you should still return the error from the internal function and handle the printing from a top-level function.

pkginfo/types.go
```diff
@@ -65,22 +65,18 @@ type Repo struct {
...
 		fmt.Sprintf("%s/%s/%s/%s", repo.URI, repo.Version, repo.Name, loc),
 		fmt.Sprintf("%s/releases/%s/%s/%s", repo.URI, repo.Version, repo.Name, loc),
 	}
 
-	var errs []error
 	for _, url := range urls {
 		_, err := helpers.CheckStatus(url)
 		if err == nil {
 			repo.URI = url
-			return nil
```
matthewrsj: So this now essentially reverses the order of precedence for picking URLs. Since you don't exit in this loop it will always pick the last url in `urls` that it finds instead of the first. Is this intentional? If it is intentional why not reverse the order of the `urls` slice instead so it is only checking the status until it finds the first one that works?

Gabib32: You're completely right, thank you. Adding a return there.

pkginfo/types.go
```diff
@@ -65,22 +65,19 @@ type Repo struct {
 	Packages []*RPM
 }
 
-func getRepoURI(repo *Repo, loc string) error {
+func getRepoURI(repo *Repo, loc string) {
```
matthewrsj: So now it is possible for `repo.URI` to remain unset after calling `getRepoURI` when there is no access to network?

matthewrsj: /me reads your commit message.

Ok, can we change the function name to something like `updateRepoURI` so it is clear that this is only updating an already-set `repo.URI` if it is able? and maybe add a comment on the function?

Gabib32: yes yes I'll do those things!

Issue #97, opened 2018-09-24
Add validation for inter-subpackage %attr mismatches


phmccarty:
Since rpmbuild permits subpackages to install the same file but with different file attributes (e.g. using the `%attr` macro), diva should flag this condition as an error, because DNF (correctly) treats this as a conflict. I will list example packages in comments that describe the conflicts.

phmccarty:
**Example**: *subpackages that specify different file permissions for the same file (`testfile` is just an arbitrary file)*:

```
Name     : test-perm-mismatch
Version  : 1
Release  : 1
URL      : https://clearlinux.org
Source0  : testfile
Summary  : Test Permissions Mismatch
License  : GPL-2.0
Requires: test-perm-mismatch-sub1

%description
Test for explicit mismatch in permissions between two files across
subpackages.

%package sub1
Summary: subpackage for test-perm-mismatch

%description sub1
subpackage for test-perm-mismatch package


%prep

%build

%install
rm -rf %{buildroot}
mkdir -p %{buildroot}/usr/share/testpkg/
cp %{SOURCE0} %{buildroot}/usr/share/testpkg/testfile


%files
%attr(0644,-,-) /usr/share/testpkg/testfile

%files sub1
%attr(0755,-,-) /usr/share/testpkg/testfile
```

phmccarty:
**Example**: *subpackages that specify different file user ownership for the same file (`testfile` is just an arbitrary file)*:

```
Name     : test-owner-mismatch
Version  : 1
Release  : 1
URL      : https://clearlinux.org
Source0  : testfile
Summary  : Test File Owner Mismatch
License  : GPL-2.0
Requires: test-owner-mismatch-sub1

%description
Test for explicit mismatch in owner between two files across
subpackages.

%package sub1
Summary: subpackage for test-owner-mismatch

%description sub1
subpackage for test-owner-mismatch package


%prep

%build

%install
rm -rf %{buildroot}
mkdir -p %{buildroot}/usr/share/testpkg/
cp %{SOURCE0} %{buildroot}/usr/share/testpkg/testfile


%files
%attr(-,0,-) /usr/share/testpkg/testfile

%files sub1
%attr(-,1000,-) /usr/share/testpkg/testfile
```

phmccarty:
**Example**: *subpackages that specify different file group ownership for the same file (`testfile` is just an arbitrary file)*:

```
Name     : test-group-mismatch
Version  : 1
Release  : 1
[...]

PR #112, opened 2018-11-07
Add file conflict check

Gabib32:
Add a check that iterates all of the package rpms from a
bundle tag, and checks that no two rpms from different
SRPMs use the same filename.


cmd/file_conflicts.go
```diff
@@ -0,0 +1,122 @@
...
+
+	bundleRPMs, err := bundleInfo.BundleDefinitions.GetAllPackages("")
+	if err != nil {
+		return r, err
+	}
+
+	allFiles := make(map[string][]*pkginfo.RPM)
+	conflicts := []string{}
+	for pkg := range bundleRPMs {
+
```
tmarcu: whitespace :stuck_out_tongue_winking_eye:

cmd/file_conflicts.go
```diff
@@ -0,0 +1,122 @@
...
+				// different SRPMs
+				for j := range allFiles[fname] {
+					if allFiles[fname][j].SRPMName != rpm.SRPMName {
+						c := fmt.Sprintf("%s and %s -> %s", rpm.SRPMName, allFiles[fname][j].SRPMName, fname)
+						conflicts = append(conflicts, c)
+					}
+				}
+				allFiles[fname] = append(allFiles[fname], rpm)
+			}
+			allFiles[fname] = []*pkginfo.RPM{rpm}
```
tmarcu: This line immediately overwrites what you did on line 111, I think you should remove one of them right?

cmd/file_conflicts.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	r.Ok(len(conflicts) == 0, "No package file conflicts")
+	for _, c := range conflicts {
+		r.Diagnostic(fmt.Sprintf("Package file conflict bettween %s", c))
```
matthewrsj: s/bettween/between/

cmd/file_conflicts.go
Gabib32: shoot. You're definitely right. It should be an if else - Great catch, thank you!

cmd/file_conflicts.go
```diff
@@ -0,0 +1,122 @@
...
+			return r, err
+		}
+		for i := range rpm.Files {
+			fname := rpm.Files[i].Name
+			// if the "file" doesn't have a hash or a symlink, it is a directory,
+			// so ignore it
+			if rpm.Files[i].Hash == "" && rpm.Files[i].SymlinkTarget == "" {
+				continue
+			}
+			if _, ok := allFiles[fname]; ok {
```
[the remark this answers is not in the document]
Gabib32: I think it would still need that loop to iterate the slice of filenames within the SRPM map, but I'll write it out really quick and see for sure.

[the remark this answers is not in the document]
Gabib32: Wait, I don't think that would make sense, because the point is to find filenames that don't have the same SRPM. If it was organized by SRPM, then we wouldn't be able to see if the same filename existed in other SRPMs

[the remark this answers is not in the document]
matthewrsj: Hmm, you are right about that. I guess this method is fine, it is just O(nm^2) at this point, where n is number of packages and m is number of files, so pretty inefficient. I'll think about it some more but this method is fine for now.

How long does this check take to run against an entire repo?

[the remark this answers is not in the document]
Gabib32: maybe a map[fname]map[SRPM]bool, so a map of a set of SRPMs? Then check if the SRPM exists, and if if does then don't worry about it?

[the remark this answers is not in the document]
matthewrsj: what about a map[fname]srpmname? then you just add new ones with srpmname as the value, and if it already exists in the map but the srpmname is a different value it is a conflict, but if it exist and srpmname matches you are good?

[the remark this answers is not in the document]
Gabib32: What about the case where there are is:
file1 -> srpm1
file2 -> srpm2
file3 -> srpm2

if we have file1[srpm1] stored,  then check file 2, its different, so we will report that. Then file3 would also be different, so then we would report again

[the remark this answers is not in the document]
matthewrsj: No that's not what I mean. It is a map[string]string so a map of filename to srpmname
```go
for _, f := range allFilesInRepo {
    if srpm, ok := fileToSRPMMap[f.Name]; ok && srpm == f.SRPMName {
        // record conflict
    }
}
```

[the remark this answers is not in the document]
matthewrsj: (made a one-character edit to the above comment that is extremely important [s/==/!=/])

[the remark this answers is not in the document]
matthewrsj: I believe we determined your way is actually faster. just recording that here.

cmd/fileconflicts.go
```diff
@@ -0,0 +1,120 @@
...
+
+	allFiles := make(map[string]string)
+	conflicts := make(map[string][]string)
+	for pkg := range bundleRPMs {
+		rpm, err := pkginfo.GetRPM(repo, pkg)
+		if err != nil {
+			return r, err
+		}
+		for i := range rpm.Files {
+			fname := rpm.Files[i].Name
```
matthewrsj: move the assignment after the `Hash` and `SymlinkTarget` check below so the variable is only created if it is needed.

cmd/fileconflicts.go
```diff
@@ -0,0 +1,120 @@
...
+					conflicts[fname] = []string{srpm, rpm.SRPMName}
+				}
+			}
+			allFiles[fname] = rpm.SRPMName
+		}
+	}
+
+	r.Ok(len(conflicts) == 0, "Package file conflicts")
+	for f, srpms := range conflicts {
+		r.Diagnostic(fmt.Sprintf("Package file conflict %s: %s", f, strings.Join(srpms, ", ")))
```
matthewrsj: This will create output like
```
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
# Package file conflict filename: packagea, packageb, packagec
```
I think we don't need to say "Package file conflict" every time and could just report something like:
```
# filename found in packages: packagea, packageb, packagec
```
which is less verbose but actually communicates more. YMMV

cmd/fileconflicts.go
```diff
@@ -0,0 +1,120 @@
...
+		for i := range rpm.Files {
+			fname := rpm.Files[i].Name
+			// if the "file" doesn't have a hash or a symlink, it is a directory,
+			// so ignore it
+			if rpm.Files[i].Hash == "" && rpm.Files[i].SymlinkTarget == "" {
+				continue
+			}
+			// a conflict exists if two rpms have the same filename, but come from
+			// different SRPMs
+			if srpm, ok := allFiles[fname]; ok && srpm != rpm.SRPMName {
```
matthewrsj: We need an additional check: if the `srpm == rpm.SRPMName` we still need to make sure they are identical files (see #97)

Gabib32: ugh I forgot about this. That sucks because now we need to store the files, or at least those 3 attr properties

matthewrsj: ah bummer you are right. I guess can't be helped though.

cmd/fileconflicts.go
```diff
@@ -0,0 +1,160 @@
...
+// fileCompareData contains the file data to compare for conflicts. The fields
+// for this struct must be exported for the deep package to use them for comparison.
+type fileCompareData struct {
+	SRPM string
+	Attrs
+}
+
+// newFileCompare creates the struct object fileCompareData with the SRPM and
+// a new Attrs struct
+func newFileCompare(f *pkginfo.File, srpm string) fileCompareData {
```
tmarcu: Is this a function we'll want to use anywhere else in the code, or just in here? (so we should export it if so)

tmarcu: :+1: thanks for clarifying.

cmd/fileconflicts.go
```diff
@@ -0,0 +1,160 @@
...
+			// a conflict exists if two rpms have the same filename, but come from
+			// different SRPMs.
+			if savedfi, ok := allFiles[fname]; ok {
+				if savedfi.SRPM != currentfi.SRPM {
+					conflictsSRPM[fname] = append(conflictsSRPM[fname], savedfi.SRPM, currentfi.SRPM)
+					// no need to compare the attrs of mismatching SRPMs
+					continue
+				}
+
+				// It is also a conflict if two rpm files have the same filename and,
```
tmarcu: rogue comma `and,`

Gabib32: ugh runaway commas :)

cmd/fileconflicts.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+var conflictFlags rpmConflictCmdFlags
+
+func init() {
+	checkCmd.AddCommand(rpmFileConflictsCmd)
+	rpmFileConflictsCmd.Flags().StringVarP(&conflictFlags.mixName, "name", "n", "clear", "name of data group")
+	rpmFileConflictsCmd.Flags().StringVarP(&conflictFlags.version, "version", "v", "0", "version to check")
+	rpmFileConflictsCmd.Flags().BoolVar(&conflictFlags.latest, "latest", false, "get the latest version from upstreamURL")
+	rpmFileConflictsCmd.Flags().BoolVar(&conflictFlags.mash, "mash", false, "pass mash to use all recent rpms, and not populate from a bundle version")
```
matthewrsj: mash isn't a widely-accepted term, just a term commonly used by the Clear team. You should use `repo` instead since this is what it is validating.

cmd/fileconflicts.go
matthewrsj: still have the rogue comma

cmd/fileconflicts.go
```diff
@@ -0,0 +1,160 @@
...
+			allFiles[fname] = currentfi
+		}
+	}
+
+	r.Ok(len(conflictsSRPM) == 0, "SRPM mismatch file conflicts")
+	for f, data := range conflictsSRPM {
+		r.Diagnostic(fmt.Sprintf("%s found in packages: %s", f, strings.Join(data, ", ")))
+	}
+
+	r.Ok(len(conflictsATTR) == 0, "%attr mismatch file conflicts")
```
matthewrsj: Do you need to escape the literal percent sign here? `%%`

Gabib32: I don't believe so. It works fine when I test it, and attempting to escape it causes an error

matthewrsj: ok u win, must just be when used with `fmt`

PR #113, opened 2018-11-07
Add download and import command

Gabib32:
previous commit implemented functionality, these are the
specific calls. Allows users to either download or import
metadata, without having to do both.

For example:
  diva download bundles -v xxxxx
  diva import rpms --latest


cmd/download.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"github.com/clearlinux/diva/diva"
+	"github.com/clearlinux/diva/internal/config"
+	"github.com/clearlinux/diva/internal/helpers"
+	"github.com/clearlinux/diva/pkginfo"
+	"github.com/spf13/cobra"
+)
+
+// the FetchingFlags struct can be found in the config file
```
matthewrsj: Instead of `config file` say `config package` since the file name could change without changing this functionality.

download/download_manifests.go
```diff
@@ -45,13 +43,24 @@ func Manifest(baseURL string, version string, component, outF string) error {
 	return nil
 }
 
+// GetMom returns the downloaded and parsed swupd.manifest mom object
+func GetMom(mInfo pkginfo.ManifestInfo) (*swupd.Manifest, error) {
+	baseCache := filepath.Join(mInfo.CacheLoc, "update")
```
matthewrsj: You don't actually use `baseCache` again after using it for `outMoM` so why not just create `outMoM` without `baseCache`?
```go
outMoM := filepath.Join(mInfo.CacheLoc, "update", mInfo.Version, "Manifest.MoM")
```

download/download_update_files.go
```diff
@@ -0,0 +1,118 @@
+// Copyright © 2018 Intel Corporation
```
matthewrsj: I'm only noticing this because you added a new file, but would it be better to make this a shorter file name? `download/update_files.go` instead of `download/download_update_files.go`? I know we also have `download/download_manifests.go` so that should probably change too. Doesn't need to be part of this PR.

download/download_update_files.go
```diff
@@ -0,0 +1,118 @@
...
+func getAllManifestFiles(mInfo pkginfo.ManifestInfo) (map[string]finfo, error) {
+	dlFiles := make(map[string]finfo)
+	baseCache := filepath.Join(mInfo.CacheLoc, "update")
+
+	mom, err := GetMom(mInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	// this is fast, no need to parallelize
```
matthewrsj: I think I originally wrote this comment right? shame on me.

download/download_update_files.go
```diff
@@ -0,0 +1,118 @@
...
+		m, err := swupd.ParseManifestFile(outMan)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, f := range m.Files {
+			// TODO: What is this and why does it cause wg.Wait() to deadlock
+			// Also, will this not download files less than MinVer (which is current version unless recursion is passed)
+			// Do we need an actual minversion that is the earliest minversion to format bump
+			// if uint(f.Version) < mInfo.MinVer ||
```
matthewrsj: some commented out code? I'm not sure if you actually wrote this code or if it is just being moved around.

download/download_update_files.go
```diff
@@ -0,0 +1,118 @@
...
+					errChan <- f.err
+				}
+			}
+		}()
+	}
+
+	for f := range dlFiles {
+		fChan <- dlFiles[f]
+	}
+	close(fChan)
```
matthewrsj:
```go
go func() {
    for e := range errCh {
        // handle error
        errs = append(errs, e)
    }
}()
```

download/download_update_files.go
```diff
@@ -0,0 +1,118 @@
...
+	wg.Wait()
+
```
matthewrsj:
```go
close(errCh)
```

download/download_update_files.go
Gabib32: Made an issue to discuss later. Not critical.

download/download_update_files.go
```diff
@@ -0,0 +1,127 @@
...
+func getAllManifestFiles(mInfo pkginfo.ManifestInfo) (map[string]finfo, error) {
+	dlFiles := make(map[string]finfo)
+	baseCache := filepath.Join(mInfo.CacheLoc, "update")
+
+	mom, err := GetMom(mInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	// "this is fast, no need to parallelize" --Matthew Johnson
```
matthewrsj: ugh.......

PR #116, opened 2018-11-09
Add ability to fetch debuginfo rpms

Gabib32:
Debuginfo RPMs live in a repo structured similarly to the
binary and source RPMs. Add an option [--debuginfo] for users
to input to the command when downloading, importing, or fetching
so that debuginfo header data can be cached and/or imported into
the database.


matthewrsj:
also needs a rebase

cmd/download.go
```diff
@@ -143,21 +144,29 @@ func runDownloadAllCmd(cmd *cobra.Command, args []string) {
 func runDownloadRepoCmd(cmd *cobra.Command, args []string) {
 	u := config.NewUinfo(downloadFlags, conf)
 
-	// by default download both SRPMs and binary RPMs, however allow the user to
-	// pass either a "binary" or "source" flag to download only one
-	if downloadFlags.BinaryRPM || (!downloadFlags.BinaryRPM && !downloadFlags.SourceRPM) {
+	// by default download both SRPMs and binary RPMs if no specific rpm option is
+	// passed. However, users can pass --source, --binary, and/or --debuginfo to
+	// choose a specific repo to download (with the ability to do more than one)
+	if downloadFlags.BinaryRPM || (!downloadFlags.BinaryRPM && !downloadFlags.SourceRPM && !downloadFlags.DebugRPM) {
```
tmarcu: This is just my opinion, but while I can read this line I'd rather not have to parse it all in my head or read the comment :) what do you think of something like:
```
noFlags := !downloadFlags.BinaryRPM && !downloadFlags.SourceRPM && !downloadFlags.DebugRPM
if downloadFlags.BinaryRPM || noFlags {
...
}
```
@matthewrsj @Gabib32  ^

tmarcu: And you can pass noFlags for all the other spots under as well & in the `fetch` one.

Gabib32: love love love :+1: :dancer:

cmd/download.go
```diff
@@ -143,21 +144,30 @@ func runDownloadAllCmd(cmd *cobra.Command, args []string) {
 func runDownloadRepoCmd(cmd *cobra.Command, args []string) {
 	u := config.NewUinfo(downloadFlags, conf)
 
-	// by default download both SRPMs and binary RPMs, however allow the user to
-	// pass either a "binary" or "source" flag to download only one
-	if downloadFlags.BinaryRPM || (!downloadFlags.BinaryRPM && !downloadFlags.SourceRPM) {
+	// by default download both SRPMs and binary RPMs if no specific rpm option is
+	// passed. However, users can pass --source, --binary, and/or --debuginfo to
+	// choose a specific repo to download (with the ability to do more than one)
+	noFlags := !fetchFlags.BinaryRPM && !fetchFlags.SourceRPM && !fetchFlags.DebugRPM
```
matthewrsj: wow this is so much better @tmarcu++

PR #117, opened 2018-11-09
Debuginfo check implementation

Gabib32:
Add two checks for the debuginfo. The first determines
whether the debug package contains files at all, and the
second determines whether the package is considered complete.


Gabib32:
This PR is only for commit 0f98112

cmd/debuginfo.go
```diff
@@ -0,0 +1,111 @@
...
+
+func init() {
+	debuginfoCmd.Flags().StringVarP(&debuginfoFlags.mixName, "name", "n", "clear", "name of data group")
+	debuginfoCmd.Flags().StringVarP(&debuginfoFlags.version, "version", "v", "0", "version to check")
+	debuginfoCmd.Flags().BoolVar(&debuginfoFlags.latest, "latest", false, "get the latest version from upstreamURL")
+	debuginfoCmd.Flags().BoolVar(&debuginfoFlags.verbose, "verbose", false, "lists failed package results")
+}
+
+var debuginfoCmd = &cobra.Command{
+	// TODO
```
matthewrsj: todo?

cmd/debuginfo.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+			if strings.HasPrefix(fname, "/usr/lib/debug") {
+				libFlag = true
+			}
+			// no need to read all files, if both lib and src files found
+			if srcFlag && libFlag {
+				break
+			}
+		}
+		if !srcFlag {
```
matthewrsj: what about `if !libFlag`? You could loop over every package file and only find source files. You would not report the missing lib files though.

cmd/debuginfo.go
```diff
@@ -0,0 +1,111 @@
...
+			missingSource = append(missingSource, pkg.Name)
+		}
+	}
+
+	r.Ok(len(emptyFiles) == 0, "Empty debuginfo files")
```
matthewrsj: What is this doing? This isn't actually checking if debuginfo files are present or not. It is checking if the package has any files at all.

cmd/debuginfo.go
Gabib32: oh I thought I had that, weird - good catch

cmd/debuginfo.go
Gabib32: Right, that is one of the checks. To get a number of how many debuginfo rpms have no files at all. It's more of a warning than a failure.

Gabib32: The package is a debug rpm

matthewrsj: ahhhhh ok

cmd/debuginfo.go
matthewrsj: image

Gabib32: HAHAHAHA

PR #121, opened 2018-11-15
bundles: check for valid bundle names

matthewrsj:
Bundle names must match the regular expression defined in this patch and
must not conflict with the reserved MoM and full names. An example is
that bundle names may not have any '.'s in their names because this
could confuse swupd-clients.

### DIFF
--- a/cmd/bundle-verify.go
+++ b/cmd/bundle-verify.go
@@ -82,6 +82,7 @@ func runVerifyBundle(cmd *cobra.Command, args []string) {
 
 	checkIncludeLoops(result, &bundleInfo)
 	checkBundleDefinitionsComplete(result, &bundleInfo)
+	checkBundleName(result, &bundleInfo)
 	checkBundleHeaderTitleMatchesFile(result, &bundleInfo)
 	checkBundleRPMs(result, &bundleInfo, &repo)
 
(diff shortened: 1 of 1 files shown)


fenrus75:
but we'd want dots in bundle names.. is the confusion real or just
potential?

matthewrsj:
The confusion is real and the example I'm thinking of is for delta and iterative manifests, which are namespaced with an appended `.D.ver` and `.I.ver`, respectively.

Pinging @otaviobp to see if this can be changed or if there are other parts of swupd-client this may cause issues

otaviobp:
Yes, this is a problem, but not just because of .D. and .I.
We also have hashed manifest names, like this
Manifest.os-core.big_hash. So if you create a bundle named os-core.big_hash you would conflict with the hashed manifests.

We always assume that we don't have '.' in bundle names in swupd, so if we are going to start using dots in bundle names we will need to do a review in swupd to make sure there's no other points of failure.

cmd/bundle-verify.go
```diff
@@ -147,6 +148,25 @@ func checkBundleDefinitionsComplete(result *diva.Results, bundleInfo *pkginfo.Bu
 	}
 }
 
+var validBundleNameRegex = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
```
Gabib32: Although the check was implemented to not allow '.' in bundle names, it seems that the regex may allow other non-valid bundle names (starting with a hyphen or underscore).

matthewrsj: are those invalid?

Gabib32: `(`^[A-Za-z0-9]+[A-Za-z0-9-_]*$`)` maybe?

matthewrsj: I mean what makes a bundle name with those leading characters invalid?

Gabib32: Is validity only determined by swupd breakage? I guess they wouldn't be invalid; maybe just not a proper convention. Sounds good though, +1

matthewrsj: It would certainly be unusual, but since it won't break anything I'd rather not make an artificial breakage by making the check more strict. I can imagine someone wanting to add their own bundles to their mix and to differentiate them from upstream mixes they may prepend the names with an underscore or hyphen (`editors` comes from upstream while `_editors` is the local mix flavor)

PR #123, opened 2018-11-21
Add mom certificate validation check

Gabib32:
Within the update content check, add a check that runs the openssl
smime command to verify the .MoM with the .MoM.sig and passed in
certpath arguement (which defaults to the Swupd_Root.pem).

### DIFF
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -1,0 +1,69 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/clearlinux/diva/diva"
+	"github.com/clearlinux/diva/internal/config"
+	"github.com/clearlinux/diva/internal/helpers"
+	"github.com/clearlinux/diva/pkginfo"
+	"github.com/spf13/cobra"
+)
+
+type certsCmdFlags struct {
+	mixName  string
+	version  string
+	latest   bool
+	certpath string
+}
+
+var certFlags certsCmdFlags
+
+func init() {
+	certsCmd.Flags().StringVarP(&certFlags.mixName, "name", "n", "clear", "name of data group")
+	certsCmd.Flags().StringVarP(&certFlags.version, "version", "v", "0", "version to check")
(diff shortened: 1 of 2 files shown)


cmd/updatecontent.go
```diff
@@ -86,6 +90,10 @@ func UCCheck(m *pkginfo.ManifestInfo) (*diva.Results, error) {
 	if err != nil {
 		return r, err
 	}
+	err = updatecontent.CheckValidCertificate(r, m.CacheLoc, m.Version, ucFlags.certpath)
```
matthewrsj: I would rather this be a separate check from the updatecontent check. This is mom sig validation which I see as different from making sure content matches manifests.

matthewrsj: (I know it was tied to the original update content check, but I don't see them as related)

updatecontent/updatecontent.go
```diff
@@ -55,6 +55,21 @@ func CheckManifestHashes(r *diva.Results, c *config.Config, version, minVer uint
 	return nil
 }
 
+// CheckValidCertificate validates the MoM is signed correctly with the ca-cert
+// Swupd_Root.pem
+func CheckValidCertificate(r *diva.Results, baseCache, version, cert string) error {
```
tmarcu: My only optional change would be to rename this to `VerifyManifest()` or `VerifyMoM()` because Validating a certificate is actually a separate process than verifying a file/signature.

cmd/certs.go
```diff
@@ -0,0 +1,69 @@
...
+	err = verifyMoM(r, mInfo.CacheLoc, mInfo.Version, certFlags.certpath)
+	helpers.FailIfErr(err)
+
+	if r.Failed > 0 {
+		os.Exit(1)
+	}
+}
+
+// verifyMoM validates the MoM is signed correctly with the ca-cert Swupd_Root.pem
+func verifyMoM(r *diva.Results, baseCache, version, cert string) error {
```
tmarcu: :+1:

matthewrsj:
As discussed offline, this is actually incomplete, as it doesn't add the command to the `check` command

PR #125, opened 2018-11-21
Use imported MoM info in UC check

Gabib32:
Change mom struct field to be MoM, and use that stored metadata for the checks instead of reading from the cache every time.

### DIFF
--- a/bloatcheck/bloatcheck.go
+++ b/bloatcheck/bloatcheck.go
@@ -64,8 +64,8 @@ func GetBundleSize(mInfo pkginfo.ManifestInfo) (map[string]int64, error) {
 
 	var bundleSizes = make(map[string]int64)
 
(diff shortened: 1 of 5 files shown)


bloatcheck/bloatcheck.go
```diff
@@ -64,8 +64,8 @@ func GetBundleSize(mInfo pkginfo.ManifestInfo) (map[string]int64, error) {
 
 	var bundleSizes = make(map[string]int64)
 
-	sort.Slice(mInfo.Mom.Files, func(i, j int) bool {
-		return mInfo.Mom.Files[i].Name < mInfo.Mom.Files[j].Name
```
matthewrsj: /me is upset he didn't notice when `MoM` changed to `Mom` the first time :(

PR #127, opened 2018-11-26
Fix cert naming to sig

Gabib32:
use the correct terminology; this is checking the validity
of the manifest.MoM.sig file, not the cert. Add check that
the manifest.MoM and manifest.MoM.sig files exist prior to
running the check.

### DIFF
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -11,42 +11,43 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type certsCmdFlags struct {
+type sigCmdFlags struct {
 	mixName  string
 	version  string
 	latest   bool
 	certpath string
 }
 
-var certFlags certsCmdFlags
+var sigFlags sigCmdFlags
 
 func init() {
-	certsCmd.Flags().StringVarP(&certFlags.mixName, "name", "n", "clear", "name of data group")
-	certsCmd.Flags().StringVarP(&certFlags.version, "version", "v", "0", "version to check")
-	certsCmd.Flags().BoolVar(&certFlags.latest, "latest", false, "get the latest version from upstreamURL")
-	certsCmd.Flags().StringVar(&certFlags.certpath, "certpath", "/usr/share/clear/update-ca/Swupd_Root.pem", "fully qualified path to ca-cert")
+	sigCmd.Flags().StringVarP(&sigFlags.mixName, "name", "n", "clear", "name of data group")
+	sigCmd.Flags().StringVarP(&sigFlags.version, "version", "v", "0", "version to check")
+	sigCmd.Flags().BoolVar(&sigFlags.latest, "latest", false, "get the latest version from upstreamURL")
+	sigCmd.Flags().StringVar(&sigFlags.certpath, "certpath", "/usr/share/clear/update-ca/Swupd_Root.pem", "fully qualified path to ca-cert")
 }
 
-var certsCmd = &cobra.Command{
-	Use:   "certs",
-	Short: "",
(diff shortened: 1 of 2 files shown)


cmd/certs.go
```diff
@@ -11,42 +12,43 @@ import (
...
+	sigCmd.Flags().StringVar(&sigFlags.certpath, "certpath", "/usr/share/clear/update-ca/Swupd_Root.pem", "fully qualified path to ca-cert")
 }
 
-var certsCmd = &cobra.Command{
-	Use:   "certs",
-	Short: "",
-	Long:  ``,
-	Run:   runCheckCerts,
+var sigCmd = &cobra.Command{
+	Use:   "sig",
```
matthewrsj: Can we call it "signature" instead? I don't really like using an abbreviation like "sig"

cmd/certs.go
```diff
@@ -61,6 +63,12 @@ func verifyMoM(r *diva.Results, baseCache, version, cert string) error {
 	mom := filepath.Join(cache, "Manifest.MoM")
 	sig := filepath.Join(cache, "Manifest.MoM.sig")
 
+	_, momExists := os.Stat(mom)
+	_, sigExists := os.Stat(sig)
+	if os.IsNotExist(momExists) || os.IsNotExist(sigExists) {
```
matthewrsj: +1 for this check, but it should be a different commit from the `certs`->`sig` change.

cmd/certs.go
```diff
@@ -61,6 +63,12 @@ func verifyMoM(r *diva.Results, baseCache, version, cert string) error {
...
+		return fmt.Errorf("manifest.MoM or manifest.MoM.sig does not exist at %s for version %s. Try running diva [download|fetch] update -v [version]", cache, version)
```
matthewrsj: can you please manually wrap this line

cmd/certs.go
```diff
@@ -62,6 +63,13 @@ func verifyMoM(r *diva.Results, baseCache, version, cert string) error {
 	mom := filepath.Join(cache, "Manifest.MoM")
 	sig := filepath.Join(cache, "Manifest.MoM.sig")
 
+	_, momExists := os.Stat(mom)
+	_, sigExists := os.Stat(sig)
+	if os.IsNotExist(momExists) || os.IsNotExist(sigExists) {
+		return fmt.Errorf(`manifest.MoM or manifest.MoM.sig does not exist at %s for version %s.
+Try running diva [download|fetch] update -v [version]`, cache, version)
```
matthewrsj: This actually adds a newline to the error string itself, which we don't really want... Also capitalize the M in Manifest.

```go
return fmt.Errorf("Manifest.MoM or Manifest.MoM.sig does not exist at %s for version %s. " +
    "Try running diva [download|fetch] update -v [version]", cache, version)
```

You know, this is an error return and error returns don't need to tell the user what to do anyways, that should be handled by the printer if at all.
```go
return fmt.Errorf("Manifest.MoM or Manifest.MoM.sig does not exist at %s for version %s.")
```

Gabib32: fiiine

Gabib32: grr error strings cannot be capitalized

matthewrsj: "the Manifest.MoM or Manifest.MoM.sig file..." maybe?

matthewrsj: Actually you can just re-word the error, since you print the filenames anyways.
```go
return fmt.Errorf("%s or %s does not exist at %s for version %s",
    mom, sig, cache, version)
```

PR #128, opened 2018-11-26
update project to use go 1.11 and modules

Gabib32:
move to the go 1.11 update of Go:
  - Use modules for dependency resolution
  - Modify the Makefile to no longer use the GOPATH (obsolete in 1.11)
  - Modify the Makefile to use the updated coverprofile arguement in checkcoverage to cover multiple packages
  - Update travis from go 1.9 to go 1.11


go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/clearlinux/diva
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	github.com/cavaliercoder/go-rpm v0.0.0-20180809195645-44d49fb2a883
+	github.com/clearlinux/mixer-tools v5.4.0+incompatible
```
phmccarty: Why is the `+incompatible` suffix used here?

matthewrsj:
You are going to need to run `go mod vendor` to get all the dependencies to work with `autospec` so you might as well do that for this PR.

go.mod
Gabib32: From my understanding after reading some of the docs is that any major version > 1.0 will have that +incompatible suffix, since mixer and redigo are v5 and v2, respectively, they are getting that suffice. As for why/what the point of that is, I'm still trying to figure out :)

Gabib32: After doing some more reading it seems we must include the specific version in the mod.go file, and the imports since it is not backwards compatible. It seems this will keep up at that specific mixer-tools version even if mixer does a bump, and we will have to manually change the major version to ensure compatibility. I may have misunderstood the docs, but I'm not sure how we want to go forward with this. 

"If the module is version v2 or higher, the major version of the module must be included in both the module path in the go.mod file (e.g., module example.com/my/mod/v2) and the package import path (e.g., import "example.com/my/mod/v2/foo")." -https://github.com/golang/go/wiki/Modules#semantic-import-versioning

PR #131, opened 2018-11-30
check: implement manifest correctness check

matthewrsj:
Add implementation for 'diva check manifests' to validate the
correctness of manifest files in relation to changes in respective RPM
repositories between two versions.

Shortcomings:
- Does not validate directory changes, as parent directories are often
  not listed with their own entry in RPMs.
- Does not validate that all deleted files in an RPM was also removed in
  repo, as there is no record in the manifest of the file type if the
  file is marked as deleted. This guard is marked with a TODO to remove
  it once this data is provided in the manifest.


manifests/check.go
```diff
@@ -0,0 +1,562 @@
...
+	toCh := make(chan *swupd.File)
+	errCh := make(chan error, workers)
+	btCh := make(chan *swupd.Manifest)
+	bfCh := make(chan *swupd.Manifest)
+
+	for i := 0; i < workers; i++ {
+		go bundleFileWorker(&wg, cLoc, momFrom, momTo, toCh, btCh, bfCh, errCh)
+	}
+
+	var fromBundles []*swupd.Manifest
```
gabibeyer: To be more optimized, make the cap of the slice the size of the bfCh
```
fromBundles := make([]swupd.Manfest, 0, len(bfCh))
```

manifests/check.go
```diff
@@ -0,0 +1,562 @@
...
+	var wg1 sync.WaitGroup
+	wg1.Add(1)
+	go func() {
+		defer wg1.Done()
+		for m := range bfCh {
+			fromBundles = append(fromBundles, m)
+		}
+	}()
+
+	var toBundles []*swupd.Manifest
```
gabibeyer: Same as fromBundles^^

Issue #82, opened 2018-08-30
check updatecontent needs fd limit


gabibeyer:
The check updatecontent command is failing while trying to open an excessive amount of files. Modify this to not allow that to happen.

matthewrsj:
I limited the concurrency quite a bit already to solve this issue. Any ideas to dynamically limit? It's tough because the go runtime doesn't know about how many fds the threads are opening, it only schedules against cpu load. Doesn't seem like a good solution to read maxfds from the filesystem either, since we can never guarantee that diva is the only thing opening files.

Open to suggestions. I don't really want to just keep decreasing parallelism as we see this issue pop up on random machines, but that may be my only choice.

Maybe some sort of mechanism that locks a mutex at a certain number of open fds? Not quite sure what that would look like though.

tmarcu:
I assume this kind of thing:
https://github.com/clearlinux/diva/blob/master/updatecontent/updatecontent.go#L55
is what's actually running out of FD's. This is entirely too many if we open one go routine per file, this should be limited to `runtime.NumCPU()` or a multiple of that at most. The rest are all clamped to numbers that won't run out right now, but e.g once we have more bundles than FD's per process that will also fail.

gabibeyer:
Fixed in PR #133 - thank you to @tmarcu  🎉

PR #133, opened 2018-12-17
Update concurrency for updatecontent

tmarcu:
We can't use a Go routine for each file because swupd.Hashcalc() calls
os.Open() on each file passed to it, which can quickly overflow the max
FD limit. This patch clamps it to the number of CPUs * 3, or the number
of files to process, whichever is smaller. Further testing should be done
if this shows to be a bottleneck.

Fixes issue #82

### DIFF
--- a/updatecontent/updatecontent.go
+++ b/updatecontent/updatecontent.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


tmarcu:
@gabibeyer I have actually no reason for choosing 3 specifically :upside_down_face: in swupd there's a few sections where 3 was an optimal multiplier, so I just went with that here and mentioned in the commit we should profile it if it's not fast enough/too much.

PR #144, opened 2019-01-10
Specify branch to pull from when getting bundles

gabibeyer:
Explicitely pull from origin master when updating the
clr-bundles repo during the download bundles process

Signed-off-by: Gabi Beyer <gabib@live.com>


gabibeyer:
not sure its necessary to have 'remote  name' be an arg too, or to just force origin.

phmccarty:
@gabibeyer Is this fixing a bug? If so, can you elaborate about it in the commit message?

In general, a `git pull` without works perfectly well when the current local branch tracks a remote branch. But if there is no tracking branch, then yes, additional arguments are needed, depending on the situation. This is why I asked if this PR is a bug fix :-) More context is needed before I can make a better suggestion.

gabibeyer:
I will update the commit. I do think there is some sort of bug with the tracking. It is producing an error; I wonder if it is my usage of git -C messing with the work-tree

gabibeyer:
So I think it is happening when I check out a specific tag, in order to get the bundles for a specific version. Putting the branch in a detached head state. I don't like the way I wrote this initially. I'm going to close this PR and fix up this logic :)

phmccarty:
@gabibeyer Ah, that makes sense. To avoid being detached, you could explore doing something like `git checkout -b <BRANCH> refs/tags/<TAG>`.

